PR #1, opened 2015-02-03
Log unknown trap OID and origin

X-dark:
Useful to alert and add unknown trap to Centreon.


garnier-quentin:
You have now a system to manage unknown traps in a distant log file.

PR #3, opened 2015-03-07
set env to path bash interpreter, because FreeBSD and other SO use difer...

lgcosta:
set ENV (/usr/bin/env) to path Bash interpreter, because FreeBSD and other SO use different path to bash -- more compatible with others SO

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #----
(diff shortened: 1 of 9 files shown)


ganoze:
@leoncx Could you merge this pull-request in the 2.7.x branch ?

ganoze:
Finally I've ported the changes against the 2.7.0-install.sh branch. See commit 01eac77689cc1584aa600243ac037847a6bc8abf.

PR #4, opened 2015-03-07
Insert commands and decision to add user/group in FreeBSD

lgcosta:


ganoze:
Thank you very much for your pull-request. It's very interesting to discover that Centreon can be used on another OS than Linux.

However I noticed that most of the diff contains space-related changes making unclear where the real changes were. Could you rework it so that we can merge it ?

PR #3757, opened 2015-08-03
Replace simple quote by space in legend

s-duret:
Replace hash by hash isn't useful
I replace simple quote by space because graph wasn't displayed with this character

### DIFF
--- a/www/class/centreonGraph.class.php
+++ b/www/class/centreonGraph.class.php
@@ -282,7 +282,7 @@ class CentreonGraph {
     */
     protected function cleanupDsNameForLegend($dsname, $reverse = false)
     {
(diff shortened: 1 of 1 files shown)


www/class/centreonGraph.class.php
```diff
@@ -282,7 +282,7 @@ protected function cleanupDsName($dsname, $reverse = false)
     */
     protected function cleanupDsNameForLegend($dsname, $reverse = false)
     {
-        $newDsName = str_replace(array("slash_", "bslash_", "pct_",  "#", "\\"), array("/", "\\", "%", "#", "\\\\"), $dsname);
```
julienmathis: Why did you remove the line 285 ?

s-duret: Because I think replacing # by # isn't useful.

julienmathis: Are you sure ? :)

PR #3769, opened 2015-08-24
Correct Centreon Broker stats display

lpinsivy:
Allow to display on each line Centreon Broker category filters as connected peers


kduret:
Be careful about indentation

PR #3879, opened 2015-11-13
Fix generateSqlLite with a non-default MySQL port

sathieu:


sathieu:
Same as #3878, on the 2.6.x branch.

PR #4289, opened 2016-03-24
Change right on clapi generation for centcore

Tpo76:
Change group to apache for centreon


Jordan91:
Hi , when an user do APPLYCFG or POLLERRESTART , Engine wasn't no restart and he must reload on centreon-web for apply configuration.

Best regards,

PR #4311, opened 2016-04-06
Use db_port in centreon scripts

sathieu:
This fixes centstorage_purge on a DB not using the standard port


sathieu:
What about other branches?

PR #4325, opened 2016-04-14
Admin are no longer viewable in ACL, needed to clean ACL

### LINKED ISSUE #4247: Admin are no longer viewable in ACL, needed to clean ACL
Hello,

Normally admin does not need to be in ACL.
But in centreon 2.6: 
- Admin users are taken into account for the count calculation
- Admin users are not visible in ACL groups configuration
- You can add an admin to an ACL group though Contact configuration

The provided patch is to display admin in ACL (as previously in Centreon).

Thanks.

smutel:
Fix #4247 

Hello,

Normally admin does not need to be in ACL.
But in centreon 2.6:

```
- Admin users are taken into account for the count calculation
- Admin users are not visible in ACL groups configuration
- You can add an admin to an ACL group though Contact configuration
```

The provided patch is to display admin in ACL (as previously in Centreon).

Thanks.

### DIFF
--- a/www/include/options/accessLists/groupsACL/formGroupConfig.php
+++ b/www/include/options/accessLists/groupsACL/formGroupConfig.php
@@ -54,7 +54,7 @@ if (($o == "c" || $o == "w") && $acl_group_id) {
     /*
      * Set Contact Childs
      */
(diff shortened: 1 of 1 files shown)


lpinsivy:
Hi

I'm not sure that to add admin in ACL is a good idea because this will complexify the  configuration displaying admin as simple user in ACL configuration.

Maybe the best way is to freeze ACL fields in contact form for admin users.

smutel:
Hello,

I am agree it is the best way.

Thanks.

Guillaume28:
For the security of all users, we can't put admins in the ACL, so i have to reject your pull request.
We thank you for your cooperation and hope to have new ideas on your part !

PR #4334, opened 2016-04-19
Fix regression from #1841

sathieu:
sso_username does not always contain an "@"


garnier-quentin:
Thanks for the feedback. i have seen it this morning. I'm working on a patch.

sathieu:
@garnier-quentin My patch doesn't fit?

garnier-quentin:
It works but i prefer to add options to do some regexp change. It will be more flexible.

garnier-quentin:
Ref 432ff2b30c86827dd9f29bda54cb753bc219046c

PR #4404, opened 2016-05-11
Update lock.pm

smutel:
Update table cron_operation correctly at the end of centreon scripts


garnier-quentin:
We have done that because of some problems. it's by design. But i'm agree we could remove/factorize  some code. But it's a too big impact on fix version. (because there is not bug).

smutel:
The issue is when you execute script /usr/share/centreon/cron/dashboardBuilder manually, the field is_running is not set to 0 at the end of the operation

garnier-quentin:
I know. It's by design. A little weird i accept. That's why it needs to do some refactoring. But it's a major impact. Maybe in 2.8.x.

PR #4415, opened 2016-05-13
Avoid displaying poller twice in poller status

smutel:
Hello,

When a poller is not running, the name of the poller is displayed twice in the poller status tooltip.

Thanks.


julienmathis:
Thanks Samuel !

PR #4502, opened 2016-06-24
Patch to avoid blank page on configuration import with prefix

tpiresgo:
Steps to reproduce : 

1. Configuration/Monitoring Engine/Load
2. Import some commands, services, service templates...
3. Redo the import with "Create new object with prefix" and choose a prefix

Observed : 
The page remains blank and there's a PHP error on HTTPD logs about the exceeded memory.


lpinsivy:
Hi,

Thank for you pull request but since we remove this menu from Centreon web.

Regards

PR #4769, opened 2016-11-09
Centreon Auto-Discovery Rules and Scan export configuration with CLAPI

pierrick-openIT:
Hello guys,

Is it possible to add Services "Auto-Discovery Rules" and "Scan" to an export with CLAPI export command ? Actually, it only export Discovery Commands.

Regards,


pierrick-openIT:
It's a mistake sorry, i created an issue.

PR #4812, opened 2016-12-02
Rights "delete" for the share views

jben94:
Add rights "delete" for the share views

New pull request, replace #4797 and #4798.

### DIFF
--- a/www/class/centreonCustomView.class.php
+++ b/www/class/centreonCustomView.class.php
@@ -126,6 +126,23 @@ class CentreonCustomView
         return true;
     }
 
+     /**
+     * Check Permission
+     * Checks if user is allowed to delete view
+     * Returns true if user can, false otherwise
+     *
+     * @param int $viewId
+     * @return bool
+     */
+    public function checkPermissionDelete($viewId)
+    {
+        $views = $this->getCustomViews();
+        if (!isset($views[$viewId]) || ($views[$viewId]['deleteable'] == 0)) {
+            return false;
+        }
+        return true;
+    }
+
     /**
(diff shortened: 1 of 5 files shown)


kduret:
I will test it soon ;)

jben94:
I will recheck, but this code is produced here.

kduret:
It does not seem to work 

Share from admin : 
image

View from my user : 
image

=> I can delete view, even if I checked "no" to "deleteable ?"

I think you can wait before new commits, I will talk about this feature with the Product Owner

jben94:
I just looked and I did not find any error in the changes.

When you say "I can delete view", did you click the button and the view was deleted for everyone, even the creator? Or is there still the "Delete view" button?

If you apply the changes of the Pull request "", maybe you will understand better.

jben94:
The "Delete view" button is there either to delete the view or to hide it (Locked: NO or YES).

lpinsivy:
Hi @jben94,

For Centreon Web 2.8.5 we rewrited users et groups permissions to have :

- Locked users
- Non locked users
- Locked groups
- Non locked groups

Locked users/groups have only the permission to add or delete an existing shared view.
Non locked users/groups can add or delete an existing shared view and modify this view.

I’m not sure it is necessary to add delete permission because all user must have permission to remove a view from their custom view dashboard. But the idea from #4801 to change text “delete” by “hide” for locked users/groups it a good idea.

Regards,

jben94:
Hi @lpinsivy ,

I do not know the new rights because I am not yet at this version, but we made these changes because it allows the unlocked user to modify the views without being able to delete it (just hide it).

PR #4819, opened 2016-12-06
Macro Path INIT.D

jben94:


lpinsivy:
Hi @jben94 

I discussed this with my colleagues and for us the best way is to use "service <processname> start/stop/restart/reload" command instead of init script file.

service command is available for operating system with systemv or systemd

jben94:
Hi @lpinsivy 

I gave this solution, to put dynamic paths in the source code.

Guillaume28:
Sorry but we have to reject tour pull request, as Ipinsivy said, we prefer use the "service" command.
But if you have any other idea for Centreon, we will have a close look to it !

jben94:
Hi @Guillaume28, 

Ok, I had given this solution because we can not use the directory /etc/init.d to put our services.
But this problem is linked to the constraints of our company.

PR #4820, opened 2016-12-06
Proposal for an option

jben94:
Hi,

I suggest you add an option in the confiuration of a poller.
This will be used to indicate whether the poller uses "sudo" in the scripts.

I suggest this because we do not use "sudo" and we can not configure it.


Guillaume28:
Hello @jben94,

We are sorry but your pull request will seriously complicate the structure of the pollers.
So we can not implement your work into Centreon.
We hope you will send us many other ideas to improve our development.

Best regards

PR #4824, opened 2016-12-08
Deploy views

jben94:
Option that allows to load the views directly in the "CustomView" of the users. 
When view sharing


lpinsivy:
@jben94 We prefer to develop a new page to administrate custom view for users. We will include this idea during development

PR #4842, opened 2016-12-19
Fix problem with Metric API : all services are sent - Now only enable…

julienmathis:
Fix problem with Metric API : all services are sent - Now only enabled services and hosts are sent.


julienmathis:
@Dacchi Possible to cherrypick ?

lpinsivy:
@Dacchi can you commit on master branch too ?

PR #4932, opened 2017-01-31
Bad path to access to alt_error.ph

lpinsivy:
If user is non-admin user, the path to access to alt_error.ph is wrong and the page is empty.


Dacchi:
Thanks for the contribution

PR #4933, opened 2017-01-31
Add possibility to copy-paste executed command ligne

lpinsivy:
Add an icon for user with ACL that allow to see executed command line to copy-paste in clipboard command without tabulation or line's return


Ridene:
For a better use of the clipboard functionality, please use a javascript library like clipboard.js to manage the copy-paste possibility.

PR #4938, opened 2017-02-01
Column 'sg_id' in field list is ambiguous

s-duret:


loiclau:
thank you for your contribution

PR #4998, opened 2017-02-17
Compatibility with old arguments

s-duret:
If we use old arguments ($ARG1$ $ARG2$ ... ) in the service configuration and one of these argument contains simple quote, so the detailed service page is blank.

In the table "services", the field "check_command" contains command_name and old arguments:
check_cluster_new!-S!'diskcluster'!20!30
The simple quote was a problem

I tested it on two platforms and it works


loiclau:
fix in #5037

Issue #4921, opened 2017-01-26
[CLAPI] Centreon 2.8.2 - Export not working when Centreon BAM 3.5.x installed


Moujimouja:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.2
**OS**: CentOS 6

**Steps to reproduce the issue:**
1. Have Centreon BAM 3.5.x installed
2. Execute an export command : /usr/bin/centreon -u admin -p xxxxxx -e > /tmp/clapi-export.txt

**Describe the results you received:**

```
/usr/bin/centreon -u admin -p xxxxxx  -e > /tmp/clapi-export.txt
PHP Fatal error:  Class '\CentreonClapi\centreonKpi' not found in /usr/share/centreon/www/class/centreon-clapi/centreonAPI.class.php on line 822
```

**Describe the results you expected:**

Export working

PR #5005, opened 2017-02-20
Can export when a module with CLAPI is installed

leoncx:
* Fix namespace problem when CLAPI object is from a module

Ref: #4921


thiuyendang:
Centreon Web 2.8.5+  Bam 3.5.x CentOs6 CentOs7
Clapi export OK

Issue #4595, opened 2016-08-12
[Performance] No more last/min/max/average


lpinsivy:
---
## BUG REPORT INFORMATION

**Centreon Web version**: Centreon Webn 2.8.0-beta1

**Steps to reproduce the issue:**
1. Go to "Monitoring > Performance"
2. Add a chart

**Describe the results you received:**

Last/Min/Max/Average value are not displayed

querwin:
The Curves menu still presents options this ones but it's never display

masgo:
Now 2.8.1 stable has been released a month ago - this bug is still here. 

Also not only is it not displayed in performance, Min/Max/Average values can no longer be seen anywhere else.

leoncx:
Link to push request #5006

masgo:
Thanks for re-adding this feature. 

There is still a little bug in the "last" values shown.

Here is an example of a temperature sensor which reports values with a precission of .1 °C. The sensor transmits its values via SNMP as integer value*1000 (21000 means 21 °C). 

These are the values shown under the Graph:
`Temperature
Last :21266.67
Min :21000.00
Max :23000.00
Average :22133.27`

While the sensor actually transmitted 21000. The perfdata is
`Temperature=21000milli°C;25000;27000;0;0`

PR #5006, opened 2017-02-20
Extend chart legend and add more information on helps

leoncx:
* Add more information on help of curve template and graph template
  * Add information for option only used for export
* Add extra information on legends and review the style
* Add action for hide or show all curve

Ref: #4595 #4083 #5033


kduret:
image

Icon missing

Guillaume28:
Everything  works fine.

masgo:
Short question: is there a (simple) way to set up a copy of my centreon (which is based on the CES packages) and then upgrade to the latest version to test the new features? 

I know how to clone my system to a VM - the question is how to upgrade from the package-based setup to the git head or something similar.

PR #5051, opened 2017-03-03
Remove "s" in popin for duration

julienmathis:


thiuyendang:
OK Centreon Web 2.8.5 (Popin visible on page Monitoring > Status Details > Services) , CentOs6 and CentOs7

PR #5070, opened 2017-03-10
Add backslashes to "Executed Check Command Line" before line breaks

sbraz:
This makes it easier to copy-paste command lines to run them from a shell.


sbraz:
I guess with the new ("copy to clipboard" feature, this isn't really needed any more…

sbraz:
This can't be merged as it is; right now this would add backslashes to the command when we click the "copy to clipboard button".

kduret:
Yes indeed :)

Furthermore, we won't add this feature to Centreon cause current command line is enough readable ;)

Issue #5029, opened 2017-02-24
[Widget] - Sharing View - pb with multiselect users


cmeschin:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.7.7

**Centreon Engine version**: 1.5.1

**Centreon Broker version**: 2.11.5

**OS**: centos 6.8

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Create view
2. add a widget
3. go to sharing view (1) and clic on user's list (2)
![widget_view1](https://cloud.githubusercontent.com/assets/8720203/23300609/9926c20a-fa86-11e6-9b76-4684dda11ca1.png)

4. select multiple user with CTRL key (3)
![widget_view2](https://cloud.githubusercontent.com/assets/8720203/23300634/b5504c4e-fa86-11e6-995e-54421fa4c80b.png)

**Describe the results you received:**
the config box disappear, only user's list stay present. At this time, it's impossible to validate your selection. (4)
![widget_view3](https://cloud.githubusercontent.com/assets/8720203/23300689/ff074586-fa86-11e6-9bc4-55edfc60e75f.png)

you must escape and restart at begening with only one user selected at each time.

**Describe the results you expected:**
the config box musn't disappear on multi select.

**Additional information you think important (e.g. issue happens only occasionally):**

lpinsivy:
Problème still present with testing verison centreon-web-2.8.5-1488975989.8ab7d7d

cgagnaire:
There still are a small bug. The horizontal scroll bars appear when the config box is closed and we hover the area : 

![image](https://cloud.githubusercontent.com/assets/23257354/26314887/a7703648-3f0f-11e7-8a67-0811969ab3eb.png)

![image](https://cloud.githubusercontent.com/assets/23257354/26314367/0f0a8a1c-3f0e-11e7-9c29-d6c4fbdf536f.png)

cgagnaire:
And the initial bug is still present.

thiuyendang:
Test 2.8.9 CentOS6 : Not OK
- "Select all" doesn't work
- Multi selection is possible but we cannot validate our choices

PR #5074, opened 2017-03-14
Can select multiple users on shared view without popin close

leoncx:
Ref: #5029


Guillaume28:
Multiple choice with shift works, but not with ctrl.

loiclau:
shift works but ctrl works randomly.

lpinsivy:
This patch doesn't work on Centreon web 2.8.10/2.8.11

Issue #4878, opened 2017-01-10
[CLAPI] Export Hostgroups - Creation of Hostgroups - HG;ADD


bmareau:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------
**Centreon Web version**: 2.7.8
**Centreon Engine version**: 1.5.2
**Centreon Broker version**: 2.11.5
**OS**: CENTOS 6.8

**Steps to reproduce the issue:**
1. Export objects : # centreon -u admin -p centreon -e > /tmp/clapi-export.txt
2. Check in file /tmp/clapi-export.txt , but object HG;ADD;xxxx;xxxx are not present

**Describe the results you received:**
When exporting the configuration via CLAPI, creation of hostgroups are not exported

**Describe the results you expected:**
When exporting the configuration via CLAPI, all objects should be exported.
The creation of hostgroups also.
And in the file we should find the creation of hostgroups ( HG;ADD;xxxx;xxxx )

Regards

lpinsivy:
OK 

...
HG;ADD;Linux;
HG;setparam;Linux;hg_activate;1
HG;addhost;Linux;Centreon-central
...

bmareau:
Hi,
Does patch work on Centreon 2.7.x ?
Regards

PR #5083, opened 2017-03-17
Centreon CLAPI export add hostgroup

leoncx:
* Fix the export of hostgroup. Add HG;ADD command

Ref: #4878


Sims24:
Thanks @leoncx great news and works like a charm here

Issue #4868, opened 2017-01-05
[CLAPI] Partial/Filtered export doesn't work as expected


Sims24:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.7.8
**Centreon Engine version**: Not relevant

**Centreon Broker version**: Not relevant

**OS**: CES 3.3

**Additional environment details (AWS, VirtualBox, physical, etc.):** VMWare VM

**Steps to reproduce the issue:**
1. /usr/bin/centreon -u admin -p admin -e --select='HOST;Centreon-Server' 
2.
3.

**Describe the results you received:**

I have in my export all the command definitions, and hosts/services that has nothing to do with my command.

**Describe the results you expected:**

I expect to have only CLAPI object and directives related to my --select

**Additional information you think important (e.g. issue happens only occasionally):**

If I had the filter like this --filter-type='^(HOST|SERVICE|HTPL|STPL|CMD)$' all informations are ok EXCEPT for CMD objects. So I think a good start (workaround) would be to force the use of --filter-type when using --select option. Also, really filter CMD object to the ones related with my Centreon-Server host and its services/templates.

I can provide you a test platform if needed. 

Thanks;

Sims24:
Thanks that works for me

Sims24:
Hi, 

After further testing, that doesn't work, also not be able to test on 2.8.x as clapi export is broken when using modules. 

Thanks,

Sims24:
Hi, 

Example of command to test : 

 /usr/bin/centreon -u admin -p centreon -e --select='HOST;Centreon-Server' --filter-type='^(HOST|SERVICE|HTPL|STPL)$'

It should export only related informations, it's not the case. You need to have some config available (several templates, EPP, ...). 

Thanks

Sims24:
As a side note, it should be the case for **every** objects. Let take a example with some sample data and a php warning that might be **one** of the root cause : 

`/usr/bin/centreon -u admin -p centreon -e --select='INSTANCE;Central' `

Error here

`PHP Warning:  Invalid argument supplied for foreach() in /usr/share/centreon/lib/Centreon/Object/Object.php on line 241`

In output I get all my instances, so the problem seems prelly : 

```
INSTANCE;ADD;Central;127.0.0.1;22;CENGINE
INSTANCE;setparam;Central;localhost;1
INSTANCE;setparam;Central;is_default;0
[...]

PR #5084, opened 2017-03-17
CLAPI export can filter for a selected host

leoncx:
* Add default filter for export with select

Ref: #4868


Sims24:
Hi @leoncx , 

https://github.com/centreon/centreon/pull/5084/commits/69792c5c4edab2585a8bbb176ead03d987e0ff63 doesn't fix the issue, the behaviour is worst. 

Before I had my host but also a lot of other that was unrelated to my export.

Now I have my host twice, one time at the top of my export, even before templating things and I still see all other hosts/templates/<objects> in my export. 

Thanks !

Guillaume28:
I approve with the remarks of Sims24, the double command the hosts.

Issue #5058, opened 2017-03-08
InfluxDB broker output config: metric columns not stored properly


tkuther:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.4

**Centreon Engine version**: 1.7.0

**Centreon Broker version**: 3.0.3

**OS**: CentOS 6.8

**Additional environment details (AWS, VirtualBox, physical, etc.):** ESX

**Steps to reproduce the issue:**
1. Go to Configuration -> Pollers -> Broker configuration
2. Add InfluxDB output somewhere
3. Enter metric colums
4. Hit save

**Describe the results you received:**
Only the last entered metric column gets saved

**Describe the results you expected:**
All metric columns get saved

**Additional information you think important (e.g. issue happens only occasionally):**
Happens always. When generating the configuration, additional metric columns are missing in the broker .xml, too.

tkuther:
Screenshot **before** saving:
![bildschirmfoto 2017-03-08 um 10 04 16](https://cloud.githubusercontent.com/assets/63717/23697126/cf80a10e-03e6-11e7-8b1c-9293ca3b526a.png)

Screenshot **after** saving:
![bildschirmfoto 2017-03-08 um 10 04 42](https://cloud.githubusercontent.com/assets/63717/23697142/dfcc3582-03e6-11e7-91cb-f3e5c4005ad6.png)

thiuyendang:
Bug reproduced on 2.8.4.
Checked on Centreon-web 2.8.6.2, CentOS6, CentOS7
Bug is fixed.

PR #5089, opened 2017-03-20
Save metrics column on InfluxDB output config

leoncx:
* Can save multiple column in InfluxDB output config

Ref: #5058


thiuyendang:
Please refer to ticket #5058

PR #5157, opened 2017-04-13
Improve performance

lpinsivy:
Apply #4894 patch from @julienmathis on 2.7.X branch


loiclau:
Broke the generation:
PHP Fatal error:  Call to a member function getPollerAclConf() on a non-object in /usr/share/centreon/www/include/configuration/configGenerate/xml/generateFiles.php on line 104

PR #5159, opened 2017-04-14
Add option for display or not the toggle all curves in views charts

leoncx:


thiuyendang:
Checked with Centreon Web 2.8.7, Widget Graph Monitoring 1.5.1.
The "i" and toggle option buttons are still enabled, only pictures are not visible anymore.

thiuyendang:
OK 2.8-9-5

PR #5219, opened 2017-05-12
[Proposal] - CDEF in curve template

jben94:
I propose to add the possibility of applying an RPN function on non-virtual curves.

Our problem was that we sometimes had incorrect values.
In the picture below, we see that we have 1.5 people connected on the equipment.

image
Or
image

After applying the CDEF 

> value,CEIL

image
Or
image


jben94:
Plusieurs points à revoir désolé, pour cette proposition trop rapide.

PR #5260, opened 2017-05-22
[Load views] Fixed database entry duplication

jben94:
If you load a view, then you remove it and then reload it. 

You can see database entry duplications for this user.


kduret:
Thank you @jben94 for this fix :)

It works for me.

But could you add a query in update script to delete duplicate entries in database ?

Thank you !

jben94:
Hi,

I have just modified the SQL update files

Issue #5254, opened 2017-05-19
Cannot modify metaservice indicator


thiuyendang:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.7.11, 2.8.8

**Centreon Engine version**: 

**Centreon Broker version**: 

**OS**: CentOS6

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Create a metaservice (Ping RTA Average for example)
2. Add indicator for this new metaservice
3. Click on the indicator record link to display the saved information of the indicator

**Describe the results you received:**
Host and service fields are disabled, we cannot modifiy them.

**Describe the results you expected:**
Ability to modify all fields for the indicator (host, service, metric...)

**Additional information you think important (e.g. issue happens only occasionally):**

thiuyendang:
OK 2.8.9-5 CentOS6

Issue #5255, opened 2017-05-19
Apply new Centreon graphical charter to add and modify pages for metaservice indicator


thiuyendang:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.7.11, 2.8.8

**OS**: CentOS6

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Create a metaservice (Ping RTA Average for example)
2. Add indicator for this new metaservice
3. Click on the indicator record link to display the saved information of the indicator

**Describe the results you received:**
For Add indicator and modify indicator pages : 
Save, reset, modify buttons have old style.
![metaservice_indic](https://cloud.githubusercontent.com/assets/25612980/26242255/07510fd6-3c88-11e7-8527-547195075fd7.png)

**Describe the results you expected:**
![newstylebuttons](https://cloud.githubusercontent.com/assets/25612980/26242443/c3104df4-3c88-11e7-90e2-c5872ff7df1b.png)

**Additional information you think important (e.g. issue happens only occasionally):**

PR #5267, opened 2017-05-23
Meta services #5254

Guillaume28:
#5254
Now we can modify indicator when we click on it (in meta services)


Guillaume28:
#5255 
Apply the actual style for the save and reset button

PR #5268, opened 2017-05-23
Recovery notification delay helpers were wrong for host and service

cgagnaire:


thiuyendang:
OK 2.8.9-6 (correction aide sur les champs Recovery notification delay pour hôte et service)

Issue #5325, opened 2017-06-08
[2.8.8] Unable to select all services in escalation form


lpinsivy:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.8

**Centreon Engine version**: N/A

**Centreon Broker version**: N/A

**OS**: N/A

**Steps to reproduce the issue:**
1. Go to "Configuration  >  Notifications  >  Escalations" menu
2. Edit or create a new escalations
3. Try to add all services into escalation with "Select all" and click on OK

**Describe the results you received:**

No services seleted

**Describe the results you expected:**

All services should be selected

thiuyendang:
OK in 2.8.10  with a limited number of services to select,  otherwise page cannot be displayed by web browser.

PR #5326, opened 2017-06-08
Check if array is empty of not, not only array exist - #5325

lpinsivy:


Dacchi:
You should use **"!empty()"** instead of **"count()"** to check if the array is not empty.
You should also factorize the multiple **"if"** statement that do the same thing

lpinsivy:
done!

Dacchi:
Thanks for the contribution

PR #5375, opened 2017-06-22
correct typo

lkappel:


kduret:
I'm on CentOS7, and the right command is : 
`service mysqld restart`

Are you using RedHat 7 ?

PR #5377, opened 2017-06-23
HostDuplicationCheck created

mariegallardo:


features/HostDuplicationCheck.feature
```diff
@@ -0,0 +1,15 @@
...
+    Background:
+        Given I am logged in a Centreon server
+        And a host is created
+
+    Scenario: Duplicate a host and check the properties
+        When I duplicate a host
+        Then the host was correctly duplicated
+        And their alias are the same
+        And their address are the same
+        And their status are the same
```
kduret: Only one step is enough to check alias, address and status
==> And their properties are the same

PR #5383, opened 2017-06-27
Optimization of configuration generation

jben94:
I propose this change to improve the speed of compilation.

Before change, for +3 800 Hosts and +32 000 services:

real    1m19.484s
user    0m32.605s
sys     0m3.982s

After change, for +3 800 Hosts and +32 000 services:

real    0m35.835s
user    0m20.714s
sys     0m1.535s

time saving : ~44 secondes


kduret:
You don't have inherited values with your query ;)

kduret:
But indeed, we have to improve that cause 44s saving is great

kduret:
I close this PR cause we are working on the linked ticket

PR #5389, opened 2017-06-28
Fix: Select All don't work on service categories

cweiland:


thiuyendang:
Checked with Centreon-Web 2.8.10-2, CentOS6, CentOS7 : OK
Reproduced with Centreon-Web 2.8.8.

PR #5392, opened 2017-06-29
Resource CFG tests Postman fixed

mariegallardo:


mariegallardo:
Just one little thing to change before merging.

PR #5396, opened 2017-06-30
#5400 Remove deprecated field : service_inherit_contact_from_host

Guillaume28:
Fix a bug about the notification with an user.
Now, the option "Inherit only contacts/contacts group from host" in the services works correctly.

### DIFF
--- a/behat.yml
+++ b/behat.yml
@@ -159,6 +159,11 @@ default:
       contexts:
       - SpecialCharactersInContactContext
 
+    contact_configuration:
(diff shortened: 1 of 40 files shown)


thiuyendang:
In 2.8.10. the column "service_inherit_contacts_from_host" has been removed from DB (DB centreon, table service) and from code by this ticket but another ticket has added the column again in DB to maintain MAP 4.2 compatibility. So, when upgrading Centreon Web from 2.8.9 to 2.8.10, the column is still present in database.

Issue #5406, opened 2017-07-04
Adding SNMP traps definition : values set to fields in Relations tab are not saved


thiuyendang:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.9

**Centreon Engine version**: 

**Centreon Broker version**: 

**OS**: CentOS6

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Add a SNMP trap definition
2. Fill Linked services and Linked service templates fields in Relations tab
3. Save

**Describe the results you received:**
The values are not saved.

**Describe the results you expected:**
The values have to be saved.

**Additional information you think important (e.g. issue happens only occasionally):**
If we fill again the fields in modify mode of the SNMP trap definition, they are saved.

PR #5417, opened 2017-07-06
Lpinsivy traps whyadmin

lpinsivy:


lpinsivy:
should be merged after PR #5406

PR #5484, opened 2017-07-31
[FIX]  ERROR: Can't find MySQL datadir

cweiland:
ps --ef | grep mysql => no datadir information
Using mysql command help


thiuyendang:
Checked with lpinsivy, 2.8.12.

PR #5522, opened 2017-08-06
Feature add : backup dedicated database

cweiland:
Instructions :
- On database server, create a user : centreon (using ssh-copy-id)
- On database server, add centreon to sudoers (to execute /usr/share/centreon/cron/centreon-backup-sql.sh)
/etc/sudoers.d/centreon
## BEGIN: CENTREON SUDO

User_Alias      CENTREON=%centreon
Defaults:CENTREON !requiretty

# backup
CENTREON   ALL = NOPASSWD: /usr/share/centreon/centreon-backup-mysql.sh *

## END: CENTREON SUDO

- copy centreon-backup-mysql.sh from central server to database server on /usr/share/centreon/

### DIFF
--- a/cron/centreon-backup.pl
+++ b/cron/centreon-backup.pl
@@ -365,7 +365,13 @@ sub databasesBackup() {
 
         if ( grep $_ == $dayOfWeek, @fullBackupDays ) {
             print "Dumping Db with LVM snapshot (full)\n";
-            `$centreon_config->{CentreonDir}cron/centreon-backup-mysql.sh -b $TEMP_DB_DIR -d $today`;
+            if($mysql_host == "127.0.0.1") {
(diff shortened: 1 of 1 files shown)


kduret:
Hi,

I cannot accept this PR cause : 

- coding style is not good
- you give fixed path for backup on distant database (/usr/share/centreon/centreon-backup-mysql.sh)
- database on 127.0.0.1 can be "localhost" too

cweiland:
Coding style : how to improve this ?
Fixed path : it's easier to do this but I'm open to any suggestion
Database on 127.0.0.1 : Ok it's my mistake

But this feature is mandatory because, for now, when a user have a distant database, "mysql dump" is the only usable method , and this is the worst method (it's very slow when the database is too large).

PR #5542, opened 2017-08-11
New feature TrapSNMPGroup

mariegallardo:


kduret:
Waiting fix on trap groups foreign key

ganoze:
Tests should be OK now. However traps associated to a group are not properly printed when editing a group (vendor is missing).

PR #5549, opened 2017-08-16
All pollers had "config changed"

s-duret:
When you deleted a host or a service, all poller had "conf changed"


thiuyendang:
Checked on 2.8.12-5, centOS6.
2 Pollers, one remoted.
Deleted one host, -> Only central poller has config changed detected.
When I modify one host or service linked to remoted poller  then config change is detected as usual.
OK.

PR #5556, opened 2017-08-18
Web post install

linceaerian:
To connect, services have to be started and the firewall turned off or allowed; Reading the post package steps is thus a need.
You could also propose the rule for the default firewall of CentOS 7 :

```
$ firewall-cmd --zone=public --add-port=80/tcp --permanent
$ firewall-cmd --reload
```

### DIFF
--- a/doc/en/installation/from_packages.rst
+++ b/doc/en/installation/from_packages.rst
@@ -62,9 +62,11 @@ Perform the command:
  ::
 
   $ yum install centreon-base-config-centreon-engine centreon
+  $ service centcore start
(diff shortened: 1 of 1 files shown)


doc/en/installation/from_packages.rst
```diff
@@ -62,9 +62,11 @@ Perform the command:
  ::
 
   $ yum install centreon-base-config-centreon-engine centreon
+  $ service centcore start
```
kduret: centcore must be start after web install to have access to database

Issue #4867, opened 2017-01-04
Problem with LDAP autoimport and groupmapping with comma in CN


itsul:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.2 

**Centreon Engine version**: 1.6.2

**Centreon Broker version**: 3.0.3

**OS**: CES 3.4

**Steps to reproduce the issue:**
1.Create a user with a comma in the CN in LDAP/ActiveDirectory
2.Configure LDAP connector with autoimport
3.Map  LDAP-Group to an ACL
4.Try to login

**Describe the results you received:**
When you try to login the following error shows up in sql-error.log:
```
syntax error QUERY : INSERT INTO contactgroup_contact_relation
    	            						(contactgroup_cg_id, contact_contact_id)
    	            					VALUES (18, )
```
It seems that this code in centreonAuth.LDAP.class.php is at least one of the problems:
```
307                 /*
308                  * Insert user in database
309                  */
310                 $query = "INSERT INTO contact
311                     (contact_template_id, contact_alias, contact_name, contact_auth_type, contact_ldap_dn, ar_id,
312                     contact_email, contact_pager, contact_oreon, contact_activate, contact_register,
313                     contact_enable_notifications)
314                                 VALUES (" . $tmplId . ", '" .
315                     htmlentities($this->contactInfos['contact_alias'], ENT_QUOTES, 'UTF-8') . "', '" .
316                     htmlentities($userDisplay, ENT_QUOTES, 'UTF-8') . "', 'ldap', '" . $userDn . "', " . $this->arId .
317                     ", " . $userEmail . ", " . $userPager . ", '1', '1', '1', '2')";
```
```
322                     /*
323                      * Get the contact_id
324                      */
325                     $query = "SELECT contact_id FROM contact
326                         WHERE contact_ldap_dn = '" . $this->pearDB->escape($userDn, false) . "'";
327                     $res = $this->pearDB->query($query);
328                     $row = $res->fetchRow();
329                     $contact_id = $row['contact_id'];
330                     $listGroup = $this->ldap->listGroupsForUser($userDn);
331                     $listGroupStr = "";
332                     foreach ($listGroup as $gName) {
333                         if ($listGroupStr != "") {
334                             $listGroupStr .= ",";
335                         }
336                         $listGroupStr .= "'" . $gName . "'";
337                     }
338                     if ($listGroupStr == "") {
339                         $listGroupStr = "''";
[...]

PR #5577, opened 2017-08-23
LDAP import with comma on master #4867

garnier-quentin:


kduret:
Property name is pearDB, but it uses PDO ;)

Issue #5266, opened 2017-05-23
[2.8.9] Multiple 'update mode' fields in massive changes


cgagnaire:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.9

**Centreon Engine version**: 1.7.3

**Centreon Broker version**: 3.0.7

**OS**: CentOS 7

**Additional environment details (AWS, VirtualBox, physical, etc.):**
VMware vSphere 6.0

**Steps to reproduce the issue:**
1. Go to 'Configuration > Hosts'
2. Select multiple hosts, and choose 'Massive Changes' from 'More actions...' drop-down menu
3. Go to the 'Notification' tab

**Describe the results you received:**
I see two 'Update mode' fields one below the other.
![image](https://cloud.githubusercontent.com/assets/23257354/26345326/dbd749b0-3fa2-11e7-99c8-81b34cf6d2c0.png)

**Describe the results you expected:**
Only one field.

**Additional information you think important (e.g. issue happens only occasionally):**
It could be even better if every "update mode" can be shown as on the 'Monitored from' field in the 'Host Configuration' tab.
![image](https://cloud.githubusercontent.com/assets/23257354/26345378/085f450a-3fa3-11e7-96be-422a52daf0e1.png)

PR #5636, opened 2017-09-06
fix(conf): Multiple 'update mode' fields in massive changes

lpinsivy:
fix(conf): Multiple 'update mode' fields in massive changes

Remove duplicate 'update mode' fields

refs: #5266


lpinsivy:
done!

PR #5655, opened 2017-09-11
feat(clapi): Add export / import acl option in clapi on master

loiclau:


www/class/centreonTopology.class.php
```diff
@@ -0,0 +1,92 @@
...
+     * @param $value
+     * @return mixed
+     * @throws Exception
+     */
+    public function getTopology($key, $value)
+    {
+        $queryTopologyPage = 'SELECT * FROM topology WHERE topology_' . $key . ' = ?';
+        $stmt = $this->db->prepare($queryTopologyPage);
+        $dbresult = $this->db->execute($stmt, array($value));
+        if (PEAR::isError($dbresult)) {
```
kduret: Do not use PEAR in centreon 2.9.x

PR #5673, opened 2017-09-14
style(monitoring): Fix style

loiclau:


www/include/monitoring/common-Func.php
```diff
@@ -187,46 +193,48 @@ function get_notified_infos_for_service($serviceId, $hostId)
...
         if ($contactAdd['contact_additive_inheritance'] == 0 && count($contacts) > 0) {
             $stopReading['contacts'] = 1;
         }
         if ($contactAdd['cg_additive_inheritance'] == 0 && count($contactGroups) > 0) {
             $stopReading['contactGroups'] = 1;
         }
-        
-        
+
+
```
ganoze: Only one \n is required.

PR #5684, opened 2017-09-20
Force ext4 FS

s-duret:


ganoze:
It seems that the error you're facing (cannot mount disk with same UUID) is likely caused by an umounted disk. So instead of forcing an UUID change by recreating the filesystem it would be better to ensure that the disk is properly unmounted before attempting to remount it.

PR welcome :-)

PR #5685, opened 2017-09-20
Force ext4 FS

s-duret:


ganoze:
See #5684.

And BTW, the current development branch (the upcoming 2.9 version) is the *master* branch.

PR #5686, opened 2017-09-20
fix(test): fix widget installation and removal tests.

ganoze:
Do not refresh widget listing page manually when installing and
removing widget as this could prevent installation or removal
from working.

### DIFF
--- a/features/bootstrap/WidgetContext.php
+++ b/features/bootstrap/WidgetContext.php
@@ -56,17 +56,12 @@ class WidgetContext extends CentreonContext
      */
     public function theWidgetIsInstalled()
     {
-        $self = $this;
         $this->spin(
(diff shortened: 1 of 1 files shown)


features/bootstrap/WidgetContext.php
```diff
@@ -75,10 +70,13 @@ function () use ($self) {
      */
     public function theWidgetIsRemoved()
     {
-        $this->page = new WidgetListingPage($this);
-        $widget = $this->page->getEntry($this->widgetName);
-        if ($widget['actions']['remove']) {
-            throw new \Exception('Widget ' . $this->widgetName . ' is not removed.');
-        }
+        $self = $this;
```
kduret: not used

PR #5742, opened 2017-10-04
[Traps] proposal to add a filter

jben94:
Hi,

I propose a new choice of filtering.

image

Because we have had the following problem: Some of our equipment, have several services using the same OID.

Example:

image

The service "BGP_Info", will trigger the good active service at the reception of a trap.

Log: 
> [1507080122] [18937] EXTERNAL COMMAND: SCHEDULE_FORCED_SVC_CHECK;ldd...;BGP_...-BDGL7826;1507080121

Regards


lpinsivy:
@garnier-quentin can you review this proposal?

PR #5746, opened 2017-10-05
Add obsess_over_hosts parameter in main centengine configuration

smutel:
The parameter obsess_over_hosts is not populated to the main configuration file of centengine.


adr-mo:
Ok parameters can now be set and saved to engine configuration file.

PR #5751, opened 2017-10-09
fix(downtime): manage DST on recurrent downtimes

kduret:


kduret:
to cherry-pick on 2.7.x and master branches

PR #5752, opened 2017-10-09
fix foreign key upgrade of traps_group table

kduret:


www/install/php/Update-2.8.14_to_2.8.15.php
```diff
@@ -38,16 +38,30 @@
...
+        );
+
+        $pearDB->query(
             "ALTER TABLE `traps_group` " .
             "  CHANGE COLUMN `traps_group_id` " .
             "  `traps_group_id` INT NOT NULL AUTO_INCREMENT"
         );
+
+        $pearDB->query(
+            "ALTER TABLE `traps_group_relation` " .
```
ganoze: Could you place the foreign key recreation after the target field got PRIMARY KEY'd and the unnecessery INDEX dropped ?

PR #5757, opened 2017-10-09
fix(acl): Add primary key and handle duplicate entries in centreon_acl

Dacchi:


p3r7:
Here is an example of a migration script.

```sql
USE `centreon_storage`;

CREATE TABLE `centreon_acl_tmp` LIKE `centreon_acl`;

INSERT INTO `centreon_acl_tmp`
SELECT * FROM `centreon_acl` AS acl
WHERE acl.service_id IS NOT NULL # leftovers from previous migration ?
GROUP BY acl.group_id, acl.host_id, acl.service_id
;

TRUNCATE `centreon_acl`;

ALTER TABLE `centreon_acl`
ADD PRIMARY KEY (`group_id`,`host_id`,`service_id`)
;

INSERT INTO `centreon_acl`
SELECT * FROM centreon_acl_tmp AS acl
GROUP BY acl.group_id, acl.host_id, acl.service_id
;

DROP TABLE centreon_acl_tmp;
```

kduret:
replaced by #6366

PR #5774, opened 2017-10-11
fix(clapi) fix duplicate htpl in stpl

loiclau:
can be merge on master


adr-mo:
OK no more duplicated lines in configuration export/import

PR #5775, opened 2017-10-12
fix(clapi) fix duplicate on stpl

loiclau:


adr-mo:
OK no more duplicated lines in configuration export/import

Issue #5458, opened 2017-07-21
Metaservices not counted properly in  statuses filter 


thiuyendang:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.10

**Centreon Engine version**: 1.7.3

**Centreon Broker version**: 3.0.8

**OS**: CentOS7

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Have different services configured + 1 metaservice on a host
2. Click on the status filter corresponding at the one of the metaservice

**Describe the results you received:**
Metaservice is in the list but the status filter displays : n-1/n records
![image](https://user-images.githubusercontent.com/25612980/28470333-b2948610-6e39-11e7-800b-4715da59b700.png)

**Describe the results you expected:**
The status filter should display n/n.

**Additional information you think important (e.g. issue happens only occasionally):**
Any service is acknowledged.
When we disable the metaservice then the status filter number is OK.

garnier-quentin:
Test and it's corrected in 2.8.11:
https://github.com/centreon/centreon/commit/79cd42d4e119a9f52b64e701f4f6bf2384e2288c

lpinsivy:
Error still present in Centreon Web 2.8.12

![screenshot at aout 30 11-15-03](https://user-images.githubusercontent.com/3351916/29864973-8acca8ba-8d74-11e7-9084-c10c72e8bf4a.png)

adr-mo:
Error fixed in Centreon Web 2.8.12-7

![error_fixed](https://user-images.githubusercontent.com/31647811/30212361-e12c053e-94a4-11e7-9545-fa31046ea504.png)

Benchmarks were done on a bench machine that provides more than 200 000 services.

Benchmarks were done using the admin account, but also an other account with ACL.

Please find the results above

![bench](https://user-images.githubusercontent.com/31647811/30212368-ea482602-94a4-11e7-858e-ad7ed22a4706.png)

PR #5703, opened 2017-09-25
fix(top counter): display too much resources

lpinsivy:
Add SELECT DISTINCT service_id from centreon_acl table

ref: #5616

PR #5782, opened 2017-10-13
Apply PR #5616 on master for top counter

lpinsivy:
fix(monitoring) : Metaservices not counted properly in statuses filter

Merge SQL requests for statuses of services to:

combine ACL in one SQL request
combine selection of total services per status and total of ACK and DT services per status
This allow to correct issue with calculation of Meta-services

refs: #5458,  #5616


loiclau:
include #5703

PR #5798, opened 2017-10-18
fix(kb): increase page limit for mediawiki migration

kduret:


thiuyendang:
OK for Hosts, Hosts templates, Services Templates
KO for Services

lpinsivy:
@thiuyendang same class for four objects, can you retry for services and check with dev?

thiuyendang:
Saw with kduret : another PR will fix the issue for services pages rename

PR #5801, opened 2017-10-20
chore(build): add 2.7.13 to upgrade path.

ganoze:


adr-mo:
Ok update went well.
     - database updated OK
     - release note OK
     - Version fields updated (login + about page) OK

Looks good to me

PR #5809, opened 2017-10-24
fix(install): properly place update to 2.8 from 2.7.

ganoze:


adr-mo:
Upgrade from 2.7.13 to 2.8.16. Database correctly updated. Looks good to me

PR #5817, opened 2017-10-25
feat(chart): add legend name when defined

cgagnaire:
It's gonna be LEGEN... wait for it... DARY!


chgautier:
Validated on 2.8.16
Legend names replace Data Source names.

Issue #5711, opened 2017-09-27
Host search not saved when activate/desactivate a host


lolokai:
1.

---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.13

**Centreon Engine version**: 1.8.0

**Centreon Broker version**: 3.0.9

**OS**: CentOS 6 - Install from RPM

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Go to Configuration - Hosts
2. Make a search
3. Activate a desactivate host

**Describe the results you received:**
The search is lost

**Describe the results you expected:**
The search should be keep

**Additional information you think important (e.g. issue happens only occasionally):**

lolokai:
Same thing when you modify the host

kadrika:
I have the same issue.

mavynet:
This is because the search form makes a POST request with 'searchH' as hostname key, but the disable/enable button makes a GET request with 'search' as the key.  

As a quick fix I replaced '$_POST["searchH"]' with '$_REQUEST["searchH"]' on lines 65, 66, 67 in /usr/share/centreon/www/include/configuration/configObject/host/listHost.php and replaced 'search' with 'searchH' on lines 272 and 274 in the same file (theese are the links of the disable/enable button)

Hope this helps until the bug is fixed

lpinsivy:
I prefer to use explicitly _GET and _POST instead of _REQUEST

```
# diff -Nru listHost.php.origin listHost.php
--- listHost.php.origin	2017-10-04 13:26:32.296048331 +0200
+++ listHost.php	2017-10-04 13:34:38.109019375 +0200
@@ -64,7 +64,9 @@

 if (isset($_POST["searchH"])) {
     $search = $_POST["searchH"];
-    $_POST["search"] = $_POST["searchH"];
+    $centreon->historySearch[$url] = $search;
+} elseif (isset($_GET["searchH"])) {
+    $search = $_GET["searchH"];
     $centreon->historySearch[$url] = $search;
 } elseif (isset($centreon->historySearch[$url])) {
     $search = $centreon->historySearch[$url];
@@ -269,9 +271,9 @@
         $selectedElements = $form->addElement('checkbox', "select[".$host['host_id']."]");

         if ($host["host_activate"]) {
[...]

PR #5827, opened 2017-10-27
Fix(configuration): Host search not saved when activate/desactivate a host

lpinsivy:
Add _GET to get host search filter
Change URL for enable/disable host

fix: #5711


adr-mo:
Issue reproduced on a 2.8.16 version of Centreon.

Applying the patch has fixed the issue.

Looks good to me.

PR #5831, opened 2017-10-27
checkIllegalChar one time only

loiclau:
for master too


adr-mo:
Code optimization -> OK
Patch applied in 2.8.x and master -> OK

Issue #5830, opened 2017-10-27
Broker remove non existing protocol


lpinsivy:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.x

**Steps to reproduce the issue:**
1. Go to Broker configuration
2. Add broker configuration or edit existing
3. Go to output

**Describe the results you received:**

You can select NDO protocol

**Describe the results you expected:**

You can't select NDO protocol

**Additional information you think important (e.g. issue happens only occasionally):**

NDO protocol is no more supported by Centreon Broker 3.0.x

PR #5832, opened 2017-10-27
fix(install): Remove non existing broker protocol

lpinsivy:
Remove occurence of NDO protocol

Fix: #5830


adr-mo:
OK no more NDO protocol in the "Serialization protocol" drop-down list in the output broker configuration page.

PR #5847, opened 2017-10-31
fix(conf): some field was not encoded

cgagnaire:
Local users password appear in admin logs.


chgautier:
Fix validated on 2.8.16
Local users' contact_passwd2 are now masked in admin logs upon modification.

Log history can still show former passwords preceding this merge though.

PR #5849, opened 2017-11-02
fix(MON)downtimeStartandStop_test issue

MatthieuMan:
getMyGTMFromUser method did not return the user GMT, return added

Refs: none


thiuyendang:
Issue reproduced in 2.8.14.
For a user whose timezone is different from Europe/Paris, set a downtime on a service or a host. In downtimelist, hours are not correct.
OK in 2.8.15

PR #5863, opened 2017-11-07
fix(security) Secure upload images

loiclau:


www/class/centreonFileManager.php
```diff
@@ -0,0 +1,175 @@
+<?php
+/**
```
kduret: Missing Centreon License header

www/class/centreonImageManager.php
```diff
@@ -0,0 +1,187 @@
+<?php
+/**
```
kduret: Missing Centreon License header

www/class/iFileManager.php
```diff
@@ -0,0 +1,15 @@
+<?php
+/**
```
kduret: Missing Centreon License header

www/class/centreonImageManager.php
```diff
@@ -0,0 +1,187 @@
...
+        parent::upload();
+        if ($insert) {
+            $img_ids[] = $this->insertImg(
+                $this->destinationPath,
+                $this->destinationDir,
+                $this->newFile,
+                $this->comment
+            );
+            return $img_ids;
+        }
```
kduret: maybe return false ?

www/class/centreonImageManager.php
```diff
@@ -0,0 +1,187 @@
...
+        } elseif ($img_info['dir_alias'] != $this->destinationDir) {
+            $old = $this->mediaPath . $img_info['dir_alias'] . '/' . $img_info["img_path"];
+            $new = $this->mediaPath . $this->destinationDir . '/' . $img_info["img_path"];
+            $this->moveImage($old, $new);
+        }
+
+        //update relation
+        $query = "UPDATE view_img_dir_relation SET dir_dir_parent_id  = '" . $dirId .
+            "' WHERE img_img_id = '" . $imgId . "'";
+        $pearDB->query($query);
```
kduret: Maybe return true ?

www/include/options/media/images/formImg.php
```diff
@@ -216,7 +220,7 @@
 }
 $action = $form->getSubmitValue("action");
 
-if ($valid) {
+if (is_array($valid)) {
```
kduret: strange validation. I think methods need to return "false" if update or upload doesn't work

adr-mo:
I've tried to upload php and js using image format (extensions and headers). Even if file is uploaded, I could not figure out how to execute it.

Everything looks good to me.

PR #5866, opened 2017-11-08
backport break background ajax #5256

loiclau:
#5256 missing backport of 2.8.10


adr-mo:
I've been doing some testing to see if the number of requests has decreased. Here are my results:

Before patch: 1 TAB = 23 Requests
Before patch: 10 TAB = 232 Requests


After patch: 1 TAB = 22 Requests
After patch: 10 TAB = 140 Requests

Looks good to me

PR #5873, opened 2017-11-09
fix(macro) host/service macro visibility

loiclau:
cherry-pick on master


adr-mo:
Before applying the patch

- I've added a custom macro to a service and defined it as a password.
- I've saved the configuration
- When going back on the service definition the password was visible and "password" checkbox was unchecked.

Applying the patch has solved the problem and the password remains a password.

Looks good to me

PR #5881, opened 2017-11-10
fix(macro) hide password macro in Executed Check Command Line

loiclau:
cherry-pick on master


adr-mo:
A custom macro defined as password for the check command is no longer visible in the status details page under "Executed Check Command Line". It is replaced by "****"

Looks good to me

PR #5884, opened 2017-11-14
fix export acl menu with comment

loiclau:


adr-mo:
Already validated with loiclau

Looks good to me

PR #5891, opened 2017-11-15
fix(acl) undefined variable host id

loiclau:


www/class/centreonACL.class.php
```diff
@@ -1594,7 +1594,7 @@ public function updateACL($data = null)
                         // Put new entries in the table with group_id
                         foreach ($groupIds as $group_id) {
                             $request2 = "INSERT INTO centreon_acl (host_id, service_id, group_id) "
-                                . "VALUES ('" . $host_id . "', NULL, " . $group_id . ")";
+                                . "VALUES ('" . $data["id"] . "', NULL, " . $group_id . ")";
```
Dacchi: J'aurais plus mis $data['id'] que $data["id"]

adr-mo:
Related to #5945

PR #5896, opened 2017-11-17
use correct id for acl host relation

loiclau:


adr-mo:
I've created a Simple User.
This Simple User is associated to a Simple Ressource ACL

I log in with this user, and create a host.

After the cron of the CentACL has been executed, the host remains displayed.

Looks good to me.

**WARNING: The user must be associated to an ACL Ressource Group in order to make it work**

PR #5898, opened 2017-11-20
Update performance.rst

MaxCentreon:
La database centreon_status n'existe plus à ce jour


adr-mo:
Ok section related to the centreon_status database has been removed.

Looks good to me

PR #5901, opened 2017-11-20
Update lifecycle in documentation

lpinsivy:
Update life cycle for all component of Centreon software suite


adr-mo:
Ok all lifecycle were updated for all the modules.

Looks good to me

PR #5903, opened 2017-11-20
show future downtime

loiclau:


adr-mo:
Future downtimes are now displayed.
No regression, current downtimes are still displayed.

Looks good to me

PR #5904, opened 2017-11-20
fix(security): prepare query and execute it

Guillaume28:
As the title said, prepare et execute the query to prevent SQL injection.


Guillaume28:
Don't accept it for the moment, small display problem.

www/include/Administration/brokerPerformance/brokerPerformance.php
```diff
@@ -252,7 +252,8 @@ function parseStatsFile($statfile)
     . "FROM cfg_centreonbroker "
     . "WHERE stats_activate='1' "
     . "AND ns_nagios_server = " . CentreonDB::escape($selectedPoller) . " ";
```
loiclau: remove php variable

thiuyendang:
OK 2.8.17-4 (reproduced in 2.8.15)

PR #5915, opened 2017-11-23
fix(mon)SQL-Vulnerability_In_CustomViews

MatthieuMan:
Update SQL query to prevent SQL injection in setRotate form

Refs: MON-2129

### DIFF
--- a/www/class/centreonUser.class.php
+++ b/www/class/centreonUser.class.php
@@ -431,26 +431,34 @@ class CentreonUser
         if (!count($parameters)) {
             return null;
         }
-
+        $queryValues = array();
         $keys = array_keys($parameters);
-
         $deleteQuery = 'DELETE FROM contact_param '
-            . 'WHERE cp_contact_id = ' . $this->user_id . ' '
-            . 'AND  cp_key IN("'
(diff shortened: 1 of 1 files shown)


www/class/centreonUser.class.php
```diff
@@ -431,26 +431,35 @@ public function setContactParameters($db, $parameters = array())
...
+        $insertQuery = 'INSERT INTO contact_param (cp_key, cp_value, cp_contact_id) VALUES (?, ?, ?)';
+        $stmt = $db->prepare($insertQuery);
         foreach ($parameters as $key => $value) {
-            if (!$first) {
-                $insertQuery .= ',';
-            }
-            $insertQuery .= '("' . $key . '","' . $value . '", ' . $this->user_id . ')';
-            $first = false;
+            $sqlParams = array($key, $value, $this->user_id);
+            $db->execute($stmt, $sqlParams);
```
kduret: We lost performance cause there is one query per parameter instead of a global query

thiuyendang:
Reproduit en 2.8.17-3
Corrigé en 2.8.17-4

PR #5783, opened 2017-10-13
API: Add HOST getparam

sathieu:
Fixes: #4343

PR #5924, opened 2017-11-27
fix(doc) doc for host get param #5783

loiclau:


chgautier:
getparam confirmed in 2.8 en and fr doc.

PR #5945, opened 2017-12-04
fix(macro) replace dot

loiclau:


chgautier:
Tested on 2.8.18-6 Centreon Web.
Dots are replaced in command line password macros.
Bug tested by creating a command containing custom macros (ex: $_SERVICEMAMACRO$“) then by 
creating a service that uses the newly created command and defining the custom macro (MAMACRO) as a password with a value including “.” and finally exporting config and checking the service detail page with the macro value replaced with “***” in the executed check command line.

PR #5946, opened 2017-12-05
Feat update host get param

loiclau:
add multi param
add missing param


adr-mo:
I'm on it

adr-mo:
Looks good to me.

All the parameters specified in the documentation can be retrieved using getparam action on hosts.
The use case where a host parameter provided does not exist is handled.

ok

PR #5948, opened 2017-12-05
fix(clapi) check illegal char in add

loiclau:


loiclau:
mon-2191

adr-mo:
I'm on it

adr-mo:
Looks good to me

I've tried to ADD a HOST using special chars from nagios_cfg table

`MariaDB [centreon]> SELECT illegal_object_name_chars from cfg_nagios;
~!$%^&amp;*&quot;|&#039;&lt;&gt;?,()= 
1 row in set (0.00 sec)
`

The special chars are not taken into account

It is the same for the other objects

PR #5952, opened 2017-12-07
fix graph filter

loiclau:
mon-2192


chgautier:
Meta Service chart display OK.
Tested on 2.8.18-4 el6

PR #5982, opened 2018-01-02
fix(meta) update old service

loiclau:


chgautier:
fix ok on latest build. status details present on meta-service.
no problems whether restarting the monitoring engine with either reload or restart methods.

Issue #5974, opened 2017-12-22
Top Counter with acl really slow


garnier-quentin:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**:  2.8.16

**Steps to reproduce the issue:**
1. Many services with ACL

**Describe the results you received:**
The request for top counter is slow with acl

```
EXPLAIN SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 ELSE 0 END) AS WARNING_TOTAL, SUM(CASE WHEN s.state = 1 AND (s.acknowledged = '1' OR s.scheduled_downtime_depth = '1') THEN 1 ELSE 0 END) AS WARNING_ACK_DT, SUM(CASE WHEN s.state = 2 THEN 1 ELSE 0 END) AS CRITICAL_TOTAL, SUM(CASE WHEN s.state = 2 AND (s.acknowledged = '1' OR s.scheduled_downtime_depth = '1') THEN 1 ELSE 0 END) AS CRITICAL_ACK_DT, SUM(CASE WHEN s.state = 3 THEN 1 ELSE 0 END) AS UNKNOWN_TOTAL, SUM(CASE WHEN s.state = 3 AND (s.acknowledged = '1' OR s.scheduled_downtime_depth = '1') THEN 1 ELSE 0 END) AS UNKNOWN_ACK_DT, SUM(CASE WHEN s.state = 4 THEN 1 ELSE 0 END) AS PENDING_TOTAL FROM hosts h, services s, instances i WHERE i.deleted = 0 AND h.enabled = 1 AND s.enabled = 1 AND i.instance_id = h.instance_id AND h.host_id = s.host_id AND (h.name NOT LIKE '_Module_%' OR h.name LIKE '_Module_Meta%') AND s.service_id IN ( SELECT DISTINCT service_id FROM centreon_acl WHERE centreon_acl.group_id IN ('33','34','14'));
+------+--------------+--------------+--------+------------------------------------+------------------+---------+--------------------------------+------+-----------------------+
| id   | select_type  | table        | type   | possible_keys                      | key              | key_len | ref                            | rows | Extra                 |
+------+--------------+--------------+--------+------------------------------------+------------------+---------+--------------------------------+------+-----------------------+
[...]

PR #5992, opened 2018-01-04
fix(sql): add index in centreon_acl for topcounter

kduret:
Refs: #5974


thiuyendang:
Installed a VM that is a clone of Master BIG in 2.8.10 version.
Duplicated several hosts  with their services. Set ACL on several non admin users (created many hosts groups and linked them to one ACL access group containing one non admin user).
Before or after upgrade, the response time for the top counter requests are similar, sometimes  even longer. I'm not convinced that the fix is efficient. But I consider that this issue is not a priority and should not prevent from releasing 2.8.18 version.

Issue #6008, opened 2018-01-17
Unreachable LDAP causes template pages timeout


kduret:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.17

**OS**: CentOS 7

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Add multiple LDAP server with wrong addresses
2. Try to go in a service template page

**Describe the results you received:**
Page is loading infinitly

**Describe the results you expected:**
Page is loaded after LDAP timeout

PR #6009, opened 2018-01-17
fix(ldap): add ldap connection timeout to avoid infinite connection

kduret:
Refs: #6008


lpinsivy:
Should be bette to define this value from LDAP menu.

cgagnaire:
Closed by newer PR #7433.

PR #6013, opened 2018-01-19
fix(conf): Propose low limit of EventMaxQueueSize to 10000

lpinsivy:
Old value was 100 000. Proposal to set lower limit to 10 000

ref: #6011


thiuyendang:
For Event Queue Max Size field in Configuration > Pollers > Broker configuration, when user types a value less than 10000 and save the form, when he goes back to the form, the value is set to 10000, that is the low limit for this field.

PR #6023, opened 2018-01-29
fix(downtime) fix host group with acl

loiclau:


chgautier:
Host group downtime fixed. Tested on 2.8.18-6.

Steps to reproduce bug:
-Deploy 2 Hosts on a version preceding 2.8.18
-Create a host group containing the deployed host (whose id would be the highest)
-Create an ACL Access group for a non admin user
-Associate the ACL Ressources (host group access) to the ACL access group
-Schedule a downtime on the host group.
The downtime is not taken into account before the fix.

PR #6027, opened 2018-01-30
Add CURLOPT_SSL_ to bypass certificate

loiclau:
#6016


chgautier:
Validated on 2.8.18-6 and MediaWiki v.1.23.9
However, bug couldn't be reproduced on 2.8.17. Lacking information if there was a specific configuration for the bug to be reproduced. For example, wiki pages could be deleted in both versions (2.8.18-6 and 2.8.17) whether the web interface was on https or not.

chgautier:
Bug reproduced on 2.8.17. No links to any mediawiki pages in the status details page in secure mode.
Links displayed on 2.8.18-6 secure mode. Fix OK

Issue #6029, opened 2018-01-31
Value of 'broker' key is emptied when page is saved


bmareau:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------
**Centreon Web version**: 2.8.17
**Centreon Engine version**: 1.8.1
**Centreon Broker version**: 3.0.11

**Steps to reproduce the issue:**
1. Go to "Administration - Parameters - Monitoring" page
2. Save without edit this page
3. In 'centreon' database > table 'options', value for 'broker' key is emptied

**Describe the results you received:**
Value is erased

**Describe the results you expected:**
Value must not be erased

Regards

PR #6033, opened 2018-02-01
fix(admin): add default value for broker option

lpinsivy:
Set "broker" value for "broker" key in order to offers possibility to use old version of Centreon DSM for example

Ref: #6029


chgautier:
Validated on 2.8.18-6.

To access broker key, type the following commands on a terminal:
mysql
use centreon
select * from options where options.key='broker';

PR #6042, opened 2018-02-05
fix(broker) use final input name.

loiclau:


thiuyendang:
In Configuration > Pollers  > Broker configuration, Output tab, Name for Lua output and section lua parameter modified : new item "Generic - stream connector" in list of values for Centreon-Broker Output and lua parameter instead of lua parameters in broker configuration file.

PR #6065, opened 2018-02-16
feat(centreonApi) add clapi set option

loiclau:
add function for export


thiuyendang:
To validate with awie module.

PR #6072, opened 2018-02-20
 feat(style) update css for response message 

loiclau:


thiuyendang:
OK for the messages style of import process but please fix the content:
- "Import successful" instead of "Import successfully" (when import process is complete)
- "Please upload a .zip archive" instead of "Please update a .zip archive" (when import file format is incorrect)

PR #6074, opened 2018-02-20
Patch clapi select filter

loiclau:
return error when object filtered does not exist


lpinsivy:
@loiclau comment should be "return error when object **filtered** does not exist", right?

chgautier:
Tested on 2.8.19-2. When a filtered object doesn't exist, the following error message shows up:
Unknown Object:  'object_type;object_name'

Issue #6085, opened 2018-02-23
after upgrade 2.8.17 to 2.8.18 


vurmil:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.18

**Centreon Engine version**: 1.8.1

**Centreon Broker version**: 3.0.13

**OS**: CES

**Additional environment details (AWS, VirtualBox, physical, etc.):** Vmware

**Steps to reproduce the issue:**
After upgrade centreon from 2.8.17 to 2.8.18
**top**
mysql %CPU **313.9**
Load Average: 15.67, 16.09, 15.94 (earlier it did not exceed 1.50)

MariaDB [(none)]> show processlist;
....
| 19416657 | centreon | localhost | centreon_storage | Query   | 8660 | Sending data | SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 EL |    0.000 |
| 19416757 | centreon | localhost | centreon_storage | Query   | 8510 | Sending data | SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 EL |    0.000 |
| 19416840 | centreon | localhost | centreon_storage | Query   | 8367 | Sending data | SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 EL |    0.000 |
| 19416852 | centreon | localhost | centreon_storage | Query   | 8358 | Sending data | SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 EL |    0.000 |
| 19416886 | centreon | localhost | centreon_storage | Query   | 8342 | Sending data | SELECT SUM(CASE WHEN s.state = 0 THEN 1 ELSE 0 END) AS OK_TOTAL, SUM(CASE WHEN s.state = 1 THEN 1 EL |    0.000 |
[...]

PR #6093, opened 2018-02-27
fix(perf): fix centreon_acl index creation on update

kduret:
Refs: #6085 #5974


thiuyendang:
2.8.19-2 When upgrading Centreon Web or when installing fresh Centreon Web then two indexes are created in table centreon_acl of DB centreon_storage.

PR #6098, opened 2018-02-28
fix(config) set default poller

loiclau:


thiuyendang:
OK when upgrading from 2.8.19 -> 2.8.20, the existing LocalHost Poller becomes the default poller.
OK for fresh install from release 2.8.20.

PR #6099, opened 2018-02-28
fix(clapi) link host with default poller if unknown poller

loiclau:
when I import a host with unknown poller by clapi


thiuyendang:
Validated in 2.8.20-5.
When exporting a host and importing it in a target environment that has another Poller than the source environment one, then the host is linled to the localhost Poller.

PR #6100, opened 2018-02-28
fix(clapi) export specific service, add host before service

loiclau:
When using CLAPI export with filter and select option, add line for service was first of add host itself


thiuyendang:
When exporting a specific service, linked host is exported too, but also other services.
cf to awie project, new issue has been created.

PR #6109, opened 2018-03-05
Fix share custom view

loiclau:


chgautier:
Fix validated on 2.8.20-el6. Shared views OK

PR #6119, opened 2018-03-08
Add Check for module

Dacchi:


www/class/centreon-clapi/centreonAPI.class.php
```diff
@@ -341,6 +347,58 @@ function ($n) {
...
+        $isValid = true;
+
+        $checkLicenseFile = _CENTREON_PATH_ . "www/modules/$moduleName/extensions/checkLicense.php";
+        if (file_exists($checkLicenseFile)) {
+            require_once $checkLicenseFile;
+        }
+
+        if (in_array($moduleName, $licensedModule)) {
+            $isValid = false;
+            $licenseFile = _CENTREON_PATH_ . "www/modules/$moduleName/license/merethis_lic_temp.zl";
```
kduret: merethis_lic_temp.zl is the right name ?

PR #6125, opened 2018-03-14
feat(clapi) add resource export filter

loiclau:


thiuyendang:
to use export clapi commands with filter on some specific resources (Poller config files, LDAP...)

Issue #6127, opened 2018-03-14
[LDAP] nested groups filter is not allowed


kduret:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.19

**OS**: CentOS 7

**Steps to reproduce the issue:**
1. Create nested groups in active directory
2. Add a user in the lower group
3. Configure LDAP to find user in the lower level of the top group : 
- user filter : `(&(samAccountName=%s)(objectClass=user)(samAccountType=805306368)(memberOf:1.2.840.113556.1.4.1941:=<top_group_dn>))`
- user group attribute : `memberOf:1.2.840.113556.1.4.1941:`
- group member attribute : `member:1.2.840.113556.1.4.1941:`

**Describe the results you received:**
if a user if imported and I reload configuration, contactgroup relations are deleted

**Describe the results you expected:**
if a user if imported and I reload configuration, the user keeps its relations

PR #6128, opened 2018-03-14
fix(ldap): allow nested groups filter in ldap configuration

kduret:
Refs: #6127


chgautier:
Validated on 2.8.20-5.
Created nested group and tested with ACLs.
Contact group relations remain after export.

PR #6129, opened 2018-03-14
fix (doc): Remote Poller documentation

chgautier:
doc fix typos remote poller EN and FR

Ref: #6086


thiuyendang:
OK, in EN doc, some format fixes have to be done later (** appear instead of bold text)

PR #6132, opened 2018-03-15
doc(fix) Missing Screenshot: Remote poller doc

chgautier:
added a screenshot present in fr doc but lacking in en doc

Ref: #6086 and #6129


thiuyendang:
OK for EN doc

PR #6136, opened 2018-03-16
Update centreonGraph.class - Display PNG content at web browser output instead of downloading it

kozfelipe:
Display PNG content at web browser output instead of downloading it, so we can use png hover url in Network Weathermap integration.


cgagnaire:
Hi @kozfelipe ,

Thanks again for your contribution.

Unfortunately after discussions the Team decided that the PR will not be merged as it breaks a feature.
We will create a ticket in our backlog to keep your need in mind, and it will maybe be implemented in future versions.

Regards.

PR #6163, opened 2018-03-22
fix(acl): manage filters (poller, host, service) on servicegroup acl

kduret:


thiuyendang:
Issue was : When SG containing several services linked to different hosts and a host category filter (linked to one single host) are applied to ACL resources access record for one user then user see all hosts whereas he should see only host linked to host category.
OK in 2.8.21

PR #6174, opened 2018-03-27
fix(api): fix php warning in realtime host api

kduret:


adr-mo:
No more php warning on specific API REST calls.

Looks good to me

PR #6141, opened 2018-03-16
2.8.19 release secfixes releasenote

uncleflo:
This branch includes the security fixes as referenced in:
https://github.com/centreon/centreon/pull/6108

This also includes the removal of the release note, as requested in:
https://github.com/centreon/centreon/pull/6108

There is a conflict with regards to the release note being added in the centreon/centreon branch, and removed in my uncleflo/centreon. Please solve this yourself.

PR #6049, opened 2018-02-07
2.8.16 for release secfixes

uncleflo:
Security fixes following an internal static scan done using HP Fortify, resolving 400 issues, minus 2.

These issues have been resolved and then tested if any improvement was not working.

Summary:
- Most changes involve adding escaping variables properly
- A file centreonEscaping.php was added, to ease escaping for future developers
- Changes are based on the idea that a malicious user _could_ get administrative access

Please add this prospective branch to the latest live branch for the next release.

PR #6175, opened 2018-03-28
secfixes

loiclau:
#6141 
#6049


www/class/centreonWidget.class.php
```diff
@@ -34,6 +34,7 @@
  */
 
 require_once _CENTREON_PATH_ . "www/class/centreonUtils.class.php";
+require_once _CENTREON_PATH_ . "www/class/esc.class.php";
```
kduret: a bit strange to have a class "esc"

PR #6189, opened 2018-04-05
enh(doc): Add chapter about how to write a stream connector

lpinsivy:


doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+
+Edit also your broker output to declare this parameter:
+
+.. image:: /_static/images/developper/lua/add_parameter.png
+   :align: center
+
+Then click **Save** and go to generate the configuration and restart **cbd**.
+
+.. note::
+   Don’t forget to restart “centengine” too to create the Centreon Broker cache.
```
bouda1: I don't know if it is useful to talk about this here. You don't use the cache here.
Maybe it will be better to talk about it when you will use it...

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+the NEB events type (real time) are linked to category 1. Moreover, the status of 
+host is link to element 14 and 24 for the status of services. To filter on these 
+events only, we will add a filter function with these properties.
+
+The new LUA script will contain:
+
+.. code-block:: lua
+
+  function init(conf)
+    logFile = conf['logFile']
```
bouda1: This can lead to questions:
Why here writing conf['logFile'] and later d.host_id ?
Maybe you could precise that conf['logFile'] == conf.logFile if the key is well structured (something like conf['Centreon is good'] could not be written as conf.Centreon...

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    broker_log:set_parameters(3, "/var/log/centreon-broker/debug.log")
+  end
+  
+  function write(d)
+    local output = ""
+  
+    local host_name = broker_cache:get_hostname(d.host_id)
+    if not host_name then
```
bouda1: Same kind of remark here: previously,, you wrote
if host_name == nil then
and now you write
if not host_name then

It is interesting for a tutorial to specify they are almost equivalent.

A variable X is false if it is not defined (X=nil) or false (X=false). Otherwise, it is always true, even X=0 is true.

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+      broker_log:info(3, "Unable to get name of host, please restart centengine")
+      host_name = d.host_id
+    end
+  
+    if d.element == 14 then
+      output = "HOST:" .. host_name .. ";" .. d.host_id .. ";" .. d.state .. ";" .. d.output
+      wrintIntoFile(output)
```
bouda1: Why is this function name not writeIntoFile() ??

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+      end
+      output = "SERVICE:" .. host_name .. ";" .. d.host_id .. ";" .. service_description .. ";" .. d.service_id .. ";" .. d.state .. ";" .. d.output
+      wrintIntoFile(output)
+      broker_log:info(output)
+    end
+    return true
+  end
+  
+  function wrintIntoFile(output)
+    local file,err = io.open(logFile, 'a')
```
bouda1: Have you notice that sometimes, you writes strings as "titi", and sometimes as 'titi' ?
In Lua, there are no performance differences between them. The only real difference is that if the string is _l'hiver_, you can write "l'hiver" and if the string is _une "bête"_, you can write 'une "bête"'.
And if you always want to use "", you can too with "une \"bête\"".
If you don't explain, it is strange to see the changes between each lines.

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    if file == nil then
```
bouda1: Why not : if not file then...

doc/fr/developper/writestreamconnector.rst
bouda1: Another Lua's specific functionality, the multiple affectation.
We can do
a,b = b,a to exchange a and b.
If you don't explain, it looks strange this affectation.
Do we only affect err as we would do in C or do we affect file and err ?

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    http_server_protocol        = "http",
+    events                      = {},
+    influx_database             = "mydb",
+    influx_user                 = "",
+    influx_password             = "",
+    max_buffer_size             = 5000,
+    max_buffer_age              = 5
+  }
+  
+  function event_queue:new(o, conf)
```
bouda1: You define an object here. Your document is a howto. And I think an object is a little complicated.
So at less, you could point to a documentation explaining how to create an object.

And here, you have to specify you want to create an object. This code is not so simple.

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    o = o or {}
```
bouda1: A Lua sugar : Explain it !

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    setmetatable(o, self)
+    self.__index = self
+    for i,v in pairs(conf) do
+      if self[i] and i ~= "events" and string.sub(i, 1, 11) ~= "__internal_" then
+        broker_log:info(1, "event_queue:new: getting parameter " .. i .. " => " .. v)
+        self[i] = v
+      else
+        broker_log:warning(1, "event_queue:new: ignoring parameter " .. i .. " => " .. v)
```
bouda1: It is the first time I see a log with a priority 1, it is interesting to make a remark about it.

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+  function event_queue:add(e)
+    local metric = e.name
+    -- time is a reserved word in influxDB so I rename it 
+    if metric == "time" then
+      metric = "_"..metric
+    end
+  
+    -- retrieve objects names instead of IDs
+    local host_name = broker_cache:get_hostname(e.host_id)
+    local service_description = broker_cache:get_service_description(e.host_id, e.service_id)
```
bouda1: Here we use the cache, don't forget to specify to restart engine to fill the cache !

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+
+  function event_queue:flush()
+    broker_log:info(2, "event_queue:flush: Concatenating all the events as one string")
+    --  we concatenate all the events
+    local http_post_data = ""
+    local http_result_body = {}
+    for i, raw_event in ipairs(self.events) do
+      http_post_data = http_post_data .. raw_event
+    end
+    broker_log:info(2, "event_queue:flush: HTTP POST request \"" .. self.http_server_protocol .. "://" .. self.http_server_address .. ":" .. self.http_server_port .. "/write?db=" .. self.influx_database .. "\"")
```
bouda1: 'event_queue:flush: HTTP POST request "' may be better...

bouda1: And "\"" can be replaced by '"'

doc/fr/developper/writestreamconnector.rst
```diff
@@ -0,0 +1,586 @@
...
+    broker_log:info(3, "event_queue:flush: HTTP POST data are: '" .. http_post_data .. "'")
+  
+    -- build url
+    local influxdb_url = self.http_server_protocol.."://"..self.http_server_address..":"..self.http_server_port.."/write?db="..self.influx_database
```
bouda1: Your coding style is strange, sometimes there are spaces around ".." and sometimes not.

lpinsivy:
@bouda1 done !

thiuyendang:
New PR will be added to complete this doc, by DBO

PR #6193, opened 2018-04-09
fix(acl): fix select all checkbox in acl actions form

kduret:


thiuyendang:
OK in Admin > ACL > Actions Access, enter a record and check/uncheck all checkboxes. Save. OK

thiuyendang:
Fixed by PR 6309

PR #6197, opened 2018-04-09
fix(style): fix style in widget preferences popin

kduret:


thiuyendang:
OK, similar behaviour in 2.8.23-2 than in 2.8.21

www/Themes/Centreon-2/style.css
```diff
@@ -1772,6 +1776,7 @@ span.state_badge {
     width: 18px !important;
     right: -10px !important;
     border-radius: 10px;
```
jdelpierre: It is not necessary to leave this line with the new line you created (border at 0)

PR #6198, opened 2018-04-10
fix(style): fix padding of buttons in custom views page

kduret:


thiuyendang:
Buttons in custom views page has changed a little bit but are similar to former ones before JQuery upgrade. They are consistent for each widget.

PR #6219, opened 2018-04-19
fix(lib): allow chaining on jquery pagination plugin

kduret:


thiuyendang:
Checked in 2.8.23-2 : scrolling from top to bottom of whole PP list. OK

Issue #5011, opened 2017-02-20
[2.8.4] restore broker configuration with clapi generate too much output and input


kermith72:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.4

**Centreon Engine version**: 1.7.0

**Centreon Broker version**: 3.0.3

**OS**: ISO Centreon 3.4 (CentOS)

**Additional environment details (AWS, VirtualBox, physical, etc.):**

**Steps to reproduce the issue:**
1. Save config with Clapi
2. Delete configuration engine, broker, poller
3. Restore config with backup

**Describe the results your results **

The central-broker-master configuration have too much input and ouput. see screenshots
![install_glpi2017-02-20_03-47-13_pm](https://cloud.githubusercontent.com/assets/12350746/23129728/1e7bc73a-f784-11e6-9665-3d6228d75c45.png)
![install_glpi2017-02-20_03-47-23_pm](https://cloud.githubusercontent.com/assets/12350746/23129735/25cc6a12-f784-11e6-9711-cfd8820efa9d.png)
![install_glpi2017-02-20_03-57-11_pm](https://cloud.githubusercontent.com/assets/12350746/23130069/5a0afcde-f785-11e6-9ae4-54f62b178d85.png)

generate a wrong config

> ?xml version="1.0" encoding="UTF-8"?  
centreonBroker  
 broker_id  ![CDATA[5]]  /broker_id  
 broker_name  ![CDATA[central-broker-master]]  /broker_name  
 poller_id  ![CDATA[2]]  /poller_id  
 poller_name  ![CDATA[Central]]  /poller_name  
 module_directory  ![CDATA[/usr/share/centreon/lib/centreon-broker]]  /module_directory  
 log_timestamp  ![CDATA[1]]  /log_timestamp  
 log_thread_id  ![CDATA[1]]  /log_thread_id  
 event_queue_max_size  ![CDATA[50000]]  /event_queue_max_size  
 command_file  ![CDATA[]]  /command_file  
 input  
  type  ![CDATA[ipv4]]  /type  
 /input  
 logger  
  type  ![CDATA[file]]  /type  
 /logger  
 output  
  type  ![CDATA[sql]]  /type  
 /output  
 input  
  name  ![CDATA[central-broker-master-input]]  /name  
 /input  
 logger  
  name  ![CDATA[/var/log/centreon-broker/central-broker-master.log]]  /name  
 /logger  
 output  
[...]

PR #6220, opened 2018-04-19
fix(clapi): fix broker endpoint ids offset

cgagnaire:
Ids start at 0 when using the web UI, were starting from 1 from CLAPI.

Refs: #5011


chgautier:
Fix OK on branch 6220-endpoint-broker-fix. Broker ids start at 0 with CLAPI.

Issue #6221, opened 2018-04-19
[2.8.22] CLAPI export uses resource macro name instead of id for setparam


cgagnaire:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 
2.8.22

**Steps to reproduce the issue:**
1. Use CLAPI export option

**Describe the results you received:**
Setparam action for RESOURCECFG object uses macro name.

**Describe the results you expected:**
Setparam action for RESOURCECFG object uses macro id.

chgautier:
Fix OK on PR #6222. Upon using the CLAPI export option, the RESOURCECFG setparam option uses the object's macro id instead of its macro name.

PR #6222, opened 2018-04-19
fix(clapi): fix resource setparam export

cgagnaire:
When importing resource macro, CLAPI is expecting macro's id to be provided with setparam action.

Refs: #6221


cgagnaire:
This fix respects the way parameter setting has been designed for this type of object.
But in the case of an import in a already populated environment, this won't work as ids will be used by other resource macro definitions.

chgautier:
Fix OK on branch 6221-setparam-resource-export-fix. The RESOURCECFG set param action uses macro id upon export.

Issue #6223, opened 2018-04-19
[2.8.22] CLAPI export uses contact name instead of alias in ACL group setcontact actions


cgagnaire:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 
2.8.22

**Steps to reproduce the issue:**
1. Use CLAPI export option
2. Import configuration in fresh install

**Describe the results you received:**
"Object not found" errors appear when setting contacts to ACL groups

**Describe the results you expected:**
No errors

**Additional information you think important (e.g. issue happens only occasionally):**
CLAPI export uses contact name instead of alias in ACL group setcontact actions

chgautier:
Fix OK on PR#6224. Upon importing the config file on a freshly installed Centreon, no errors occur concerning ACL contact groups.

PR #6224, opened 2018-04-19
fix(clapi): fix acl group setcontact export

cgagnaire:
When importing ACL groups, CLAPI is expecting contact alias to be provided with setcontact action.

Refs: #6223


chgautier:
Fix OK on branch 6224-acl-group-setcontact-export-fix. Upon importing ACL contacts, there aren't any errors.

PR #6230, opened 2018-04-23
Add API Command for Host getinstance

richardlawley:
This PR adds the corresponding "getinstance" command to go with "setinstance" on Host.

```
./centreon -u admin -p password -o HOST -a getinstance -v "TEST-HOST"
1;Central
```

Documentation added.

PR #6235, opened 2018-04-24
API Create function show_poller

Xav-Deb:
Function for request which poller by host : 
curl -s -k -H centreon-auth-token:$TOKEN= -H Content-Type:application/json -d '{"action":"show_poller","object":"host","values":"hostname"}' 'https://centreon_server/centreon/api/index.php?action=action&object=centreon_clapi'

### DIFF
--- a/www/class/centreon-clapi/centreonHost.class.php
+++ b/www/class/centreon-clapi/centreonHost.class.php
@@ -323,6 +323,39 @@ class CentreonHost extends CentreonObject
     }
 
     /**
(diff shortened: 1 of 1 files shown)


lpinsivy:
Can you extend associated documentation?

richardlawley:
I submitted PR #6230 the day before this it would seem with the same functionality, but called getinstance (since the command to change poller is setinstance).

loiclau:
thanks for the participation,I have just change some things #7199

PR #6250, opened 2018-04-27
fix(sec): Fix SQL injection in dashboard

leoncx:
* Fix some SQL Injections

### DIFF
--- a/www/include/reporting/dashboard/xmlInformations/GetXmlHost.php
+++ b/www/include/reporting/dashboard/xmlInformations/GetXmlHost.php
@@ -3,34 +3,34 @@
  * Copyright 2005-2016 Centreon
  * Centreon is developped by : Julien Mathis and Romain Le Merlus under
  * GPL Licence 2.0.
- * 
- * This program is free software; you can redistribute it and/or modify it under 
- * the terms of the GNU General Public License as published by the Free Software 
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
  * Foundation ; either version 2 of the License.
- * 
+ *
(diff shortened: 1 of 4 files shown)


www/include/reporting/dashboard/xmlInformations/GetXmlHost.php
```diff
@@ -3,34 +3,34 @@
  * Copyright 2005-2016 Centreon
```
sc979: 2018 ;)

www/include/reporting/dashboard/xmlInformations/GetXmlHost.php
```diff
@@ -61,9 +70,9 @@
 
     if ($accessHost) {
         $DBRESULT = $pearDBO->query(
-            "SELECT  * FROM `log_archive_host` WHERE host_id = "
+            "SELECT  * FROM `log_archive_host` WHERE host_id = '"
```
kduret: Is it a good idea to update this query like that ? Because it's not working with mysql strict mode

www/include/reporting/dashboard/xmlInformations/GetXmlService.php
```diff
@@ -61,8 +70,8 @@
     if ($accessService) {
         $DBRESULT = $pearDBO->query(
             "SELECT  * FROM `log_archive_service` WHERE host_id = '".
-            $pearDBO->escape($_GET["host_id"])."' AND service_id = ".
-            $pearDBO->escape($_GET["id"])." ORDER BY `date_start` DESC"
+            $pearDBO->escape($_GET["host_id"])."' AND service_id = '".
```
kduret: Is it a good idea to update this query like that ? Because it's not working with mysql strict mode

PR #6254, opened 2018-04-27
fix(jQuery): fix broken input in reporting_dashboard

sc979:
correct the broken input in jQuery


thiuyendang:
OK Reporting > Dashboard. Input Form has changed a little bit for the style but behaviour is OK (export csv OK too).

PR #6267, opened 2018-05-03
CP7M2 migrate quickform to composer

v-radev:
Remove all local QuickForm files and replace all usages with a composer package. Manually load custom _select2_ and _advmultiselect_ fields.


www/include/Administration/parameters/engine/form.php
```diff
@@ -133,8 +133,8 @@
 $form->registerRule('is_writable_file_if_exist', 'callback', 'is_writable_file_if_exist');
 $form->registerRule('isNum', 'callback', 'isNum');
 
-$form->addRule('nagios_path_img', _("The directory isn't valid"), 'is_valid_path_images');
-$form->addRule('nagios_path', _("The directory isn't valid"), 'is_valid_path');
+// $form->addRule('nagios_path_img', _("The directory isn't valid"), 'is_valid_path_images'); - Field is not added so no need for rule
```
kduret: If you see dead code, delete it instead of comment it

PR #6268, opened 2018-05-03
refactor(base): replace deprecated Archive_Zip class

vhr:
replace deprecated Archive_Zip with ZipArchive class from PHP ZIP module and add composer dependency for Archive_Tar

Resolves #CP7M-13


kduret:
Could you use PharData to manage tar files ?
It should be nice to avoid pear dependency

vhr:
@kduret I replace pear Archive_Tar package with PharData

PR #6274, opened 2018-05-08
Use group with template

OpensourceMonitoringRepository:
Its easyer to have host groups on host template. 
We have host categories parents on template but we dont have filter on monitoring view to use this.
The best for me its to define host group from template to have filter on monitoring view
We can Add host group with massive change but we cant use massive change to remove one host group for all selected host
Its the same with services.
this feature has disappeared from centeron.
thank you to accepte this pull request


lpinsivy:
Hi @OpensourceMonitoringRepository 

Can you rebase your PR because this one include to many commit.

Regards,

lpinsivy:
Can you remake your PR from a branch based on 2.8.x please

PR #6289, opened 2018-05-18
Prevent permission denied centcore cmd for downtimemanager

lpinsivy:
Try to prevent error in ​downtimeManager.log by create an dedicated file for Centcore each time the cron needs to send external commands:

`PHP Warning:  file_put_contents(/var/lib/centreon/centcore.cmd): failed to open stream: Permission denied in /usr/share/centreon/www/class/centreonDowntime.Broker.class.php on line 505`


adr-mo:
OK I managed to reproduce the problem, and the fix provided works.

Looks good to me.

PR #6290, opened 2018-05-18
Sort by service name after status in service grid

richardlawley:
Minor change, sorts services in the "Services Grid" view by service name after status.  Currently they are sorted by status only, which means that two hosts with the same services sometimes get displayed in different orders.

### DIFF
--- a/www/class/centreonMonitoring.class.php
+++ b/www/class/centreonMonitoring.class.php
@@ -167,7 +167,7 @@ class CentreonMonitoring
                 . " ";
         }
 
-        $rq .= " order by tri asc";
+        $rq .= " order by tri asc, service_name";
         
         $tab = array();
         $DBRESULT = $objXMLBG->DBC->query($rq);


thiuyendang:
Hello Rich ! Please, could you tell me exactly in which Centreon's page you expect to see a change in the sort order ?
Is it in Monitoring > Status details view > Services or in the Custom views for the Service monitoring widget ?
In Monitoring > Status details view > Services : currently, services seem to be sorted by descending host name and ascending service name, 
In service monitoring widget : ascending host name and ascending service name
I have integrated your commit in my testing environment and it doesn't change anything...

richardlawley:
It's Monitoring > Status Details > Services Grid, which I think also affects some of the Widgets for Custom Views (I use hostgroup-monitoring).  It may be tricky for you to reproduce, as without the sort order its down to the default sorting provided by the database, and it's entirely possible that you'll have two hosts with the same order of services.  However, as the screenshot below shows (of two identical hosts), sometimes the services come back in a different order:

image

thiuyendang:
OK with branch 6290-service-grid-sortorder 
--
Before fix:
2018-06-21_10h55_48
After fix:
2018-06-19_13h35_45

PR #6296, opened 2018-05-22
Move writeIntoFile function

s-duret:
We need to defined this function before to call it


thiuyendang:
OK with branch 6296-move-WriteIntoFile-function

PR #6313, opened 2018-05-29
Cp7 m510 centreon awie (fixes)

vhr:


www/class/centreon-clapi/centreonObject.class.php
```diff
@@ -347,9 +348,10 @@ public function setparam($objectId, $params = array())
      * Shows list
      *
      * @param array $params
+     * @param array $filters
      * @return void
      */
-    public function show($params = array(), $filters = array())
+    public function show(array $params = NULL, array $filters = NULL)
```
kduret: set NULL in lower case

Issue #5627, opened 2017-09-05
Bad Synchronization between LDAP and Centreon


Crash-Fox:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.11

**Centreon Engine version**: 1.7.2

**Centreon Broker version**: 3.0.8

**OS**: CentOS 6.9

**Additional environment details (AWS, VirtualBox, physical, etc.):** VM

**Steps to reproduce the issue:**
1. Activate the option "Sauvegarde du mot de passe LDAP" for the LDAP source in the menu "Administration->LDAP->LDAP_Source".
2. Users changes there AD passwords.
3. Desactivate the option "Sauvegarde ..." for the LDAP source.
4. Users try to connect with to connect to Centreon Web using there new passwords.

**Describe the results you received:**
Users try to connect to Centreon Web using there new passwords but they can't. The LDAP link between the AD and Centreon doesn't work when the password of an AD user is stocked into the database of Centreon.

**Describe the results you expected:**
Users should succeed to connect to the web interface of Centreon using there new AD passwords.

Crash-Fox:
Is it possible to implement the correction of this issue in the version 2.9.0 of Centreon. This problem really impact our supervision system.

lpinsivy:
All bugs are prioritized as follows:

1. priority/blocker: next sprint (next bugfix release)
2. priority/critical: next sprint (next bugfix release)
3. priority/major: next sprint if possible, else following sprint (regarding priority/blocker and priority/critical)

Crash-Fox:
Thanks for your response. Do you have an idea when the 2.9.0 version of Centreon will be released ?

chgautier:
Fixed with PR #6347 on branch MON-2725-ldap-passwords. Using LDAPAdmin to change passwords, tested different types of passwords comprised of lowercase and uppercase letters, symbols and numbers. FIX OK.  After taking out the store password option, users can log in with their new AD passwords.

PR #6347, opened 2018-06-05
fix(ldap): remove contact password if ldap password storage is disabled

kduret:
Refs: #5627


chgautier:
Using LDAPAdmin to change passwords, tested different types of passwords comprised of lowercase and uppercase letters, symbols and numbers. FIX OK. After taking out the store password option, users can log in with their new AD passwords.

Issue #5294, opened 2017-06-01
[CLAPI] Partial / Filtered export does not work as expected for HC, SC, CG


Sims24:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**:  2.7.11 

**Centreon Engine version**:  NR 

**Centreon Broker version**: NR

**OS**: EMS

**Additional environment details (AWS, VirtualBox, physical, etc.):** NR

**Steps to reproduce the issue:**
1. /usr/bin/centreon -u admin -p password --select='SC;Disk' --filter-type 'SC' -e

**Describe the results you received:**
All my service categories being exported 

**Describe the results you expected:**
Only category disk and members being exported

**Additional information you think important (e.g. issue happens only occasionally):**
- Also present for HC / CG Host categories / Contact groups objects
- Need similar fix as https://github.com/centreon/centreon/issues/4868
- Also, side effect of fixes linked with issue above is that I had to define --filter-type to have a output for several object, that is not a big issue may be we could set some kind of 'default' regarding object in --select and overwrite if options is setted.

PR #6355, opened 2018-06-07
refacto(clapi): refacto clapi export

kduret:
ST export ==> does not export HT, only HT relations 
SVC export ==>does not export host, only host relations
HG ==> does not export linked hosts, only relations
SG ==> does not export linked services, only relations

Refs: #5294


thiuyendang:
OK with branch clapi_filter

PR #6360, opened 2018-06-08
fix(api): API and CLAPI testing

vhr:
disable all PHP errors in CLAPI (bin/centreon)
add interface and trait insteed checking for method
add missing arguments in Factory constructor
remove loading of centreon.conf.php when bootstrap.php is in use
add moduleId in centreon_administration_module when use methods for upgrading and removing a module
fix undefined variable in centreonDependency.class.php
fix undefined offset in centreonHostGroup.class.php
use sendResult instead sendJson in class CentreonWebService
add default status if missing in traps list

Resolves CP7M-64

### DIFF
--- a/config/centreon.config.php
+++ b/config/centreon.config.php
@@ -55,4 +55,4 @@ if (file_exists(_CENTREON_ETC_ . '/centreon.conf.php')) {
     }
 }
 
-$classdir = _CENTREON_PATH_.'/www/class';
+$classdir = _CENTREON_PATH_ . '/www/class';
--- a/config/centreon.config.php
+++ b/config/centreon.config.php
@@ -10,6 +10,7 @@ $constants = array(
     '_CENTREON_LOG_' => '@CENTREON_LOG@',
     '_CENTREON_VARLIB_' => '@CENTREON_VARLIB@'
 );
+
 foreach ($constants as $name => $value) {
     if (!defined($name)) {
         define($name, $value);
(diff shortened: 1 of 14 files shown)


config/centreon.config.php
```diff
@@ -19,17 +20,39 @@
...
     if (isset($dependencyInjector)) {
         $dependencyInjector['configuration'] = function ($c) use ($conf_centreon) {
             return new CentreonLegacy\Core\Configuration\Configuration($conf_centreon);
         };
     }
-}
\ No newline at end of file
+}
+
+$classdir = _CENTREON_PATH_.'/www/class';
```
kduret: missing spaces around dot

www/api/class/centreon_administration_module.class.php
```diff
@@ -72,11 +68,17 @@ public function postUpgrade()
...
             throw new \Exception('Missing argument : name');
         } else {
-            $name = $this->arguments['name'];
+            $moduleName = $this->arguments['name'];
+        }
+
+        $moduleId = $this->getModuleId($moduleName);
+
+        if (!$moduleId) {
+            throw new \Exception('The module not installed');
```
kduret: "The module is not installed"

www/api/class/centreon_administration_module.class.php
```diff
@@ -90,12 +92,43 @@ public function postRemove()
...
```
kduret: "The module is not installed"

www/api/class/centreon_administration_module.class.php
```diff
@@ -90,12 +92,43 @@ public function postRemove()
...
+    private function getModuleId($moduleName)
+    {
+        $sql = 'SELECT id FROM modules_informations WHERE name = :name';
+        $params = [
+            'name' => $moduleName,
+        ];
+
+        $result = $this->dependencyInjector['configuration_db']->query($sql, $params);
+
+        $row = $result->fetchRow();
```
kduret: use fetch instead of fetchRow (we will delete fetchRow method later)

PR #6361, opened 2018-06-08
fix(knowledgebase): fix compatibility with MW1.30

victorvassilev:
Apply new API rules and add compatibility downwards to 1.27

Resolves: CP7M-63

### DIFF
--- a/www/class/centreon-knowledge/wikiApi.class.php
+++ b/www/class/centreon-knowledge/wikiApi.class.php
@@ -72,10 +72,12 @@ class WikiApi
     private function getCurl()
     {
         $curl = curl_init();
-
+        $cookiefile = tempnam("/tmp", "CURLCOOKIE");
         curl_setopt($curl, CURLOPT_URL, $this->url);
         curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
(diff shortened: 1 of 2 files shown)


www/class/centreon-knowledge/wikiApi.class.php
```diff
@@ -72,10 +72,12 @@ public function __construct()
     private function getCurl()
     {
         $curl = curl_init();
-
+        $cookiefile = tempnam("/tmp", "CURLCOOKIE");
         curl_setopt($curl, CURLOPT_URL, $this->url);
         curl_setopt($curl, CURLOPT_RETURNTRANSFER, true);
         curl_setopt($curl, CURLOPT_POST, true);
+        curl_setopt($curl, CURLOPT_COOKIEFILE, $cookiefile);
```
sc979: can you add an expiration time using for example `CURLOPT_CONNECTTIMEOUT`

victorvassilev: added removal of cookies as agreed

sc979: thanks

Issue #6080, opened 2018-02-22
[2.8.18] API centreon_realtime_services object status filter


kozfelipe:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.18

**Centreon Engine version**: 1.8.1

**Centreon Broker version**: 3.0.13

**OS**: 2.6.32-696.20.1.el6.x86_64

**Steps to reproduce the issue:**
1. consult centreon_realtime_services objects with action list and any status filter 

**Describe the results you received:**
Error, bad status parameter

**Describe the results you expected:**
services list with status filter 

**Additional information you think important (e.g. issue happens only occasionally):**
I got bad status parameter after upgrading from 2.8.17 to 2.8.18

Archie1978:
Hello,
     do you know when the problem will be fixed? We have broken communications since 3 months :/

edhil:
Hello, 
I have the same issue on centreon 2.8.23

edhil:
Seems like a cut&paste bug between centreon_realtime_hosts.class.php and centreon_realtime_services.class.php
Here is a quick patch for centreon_realtime_services.class.php : 
`diff centreon_realtime_services.class.php.ori centreon_realtime_services.class.php 
126c126,127
<             $statusList = array('up', 'down', 'unreachable', 'pending', 'all');
---
>             $statusList = array('ok','warning','critical','unknown','pending','all');
> 
`

PR #6363, opened 2018-06-08
fixe (api) #6080

edhil:
small patch to probable cut/paste bug


chgautier:
Fix OK on branch 6363-api-fix. Filters have been updated.

PR #6366, opened 2018-06-13
fix(DB): upgrade DB to correct a bug

sc979:
Removing MUL key with PRI key
Avoiding to launch the patch when customers service already applied it


www/install/php/Update-2.8.23_to_2.8.24.php
```diff
@@ -0,0 +1,88 @@
...
+    /**
+     * Stopping the cron
+     */
+    shell_exec('sudo service crond stop');
+    sleep(10);
+
+    /**
+    * Retrieving index occurrences and dropping keys
+    */
+    $search_index = 'SHOW INDEX FROM centreon_storage.centreon_acl WHERE Key_name LIKE \'index%\'';
```
kduret: do not use "centreon_storage" name directly. Some people have custom db name.
If you use pearDBO variable, you use storage database, so you don't need to specify it.

www/install/php/Update-2.8.23_to_2.8.24.php
```diff
@@ -0,0 +1,88 @@
...
+        $needUpgrade = "true";
+    }
+}
+
+if ($needUpgrade === "true") {
+
+    /**
+     * Stopping the cron
+     */
+    shell_exec('sudo service crond stop');
```
kduret: do not stop service in upgrade process.
You have to implement a system to stop temporarily centAcl cron

PR #6370, opened 2018-06-15
Cp7 m 69 fix ldap groups

victorvassilev:


www/class/centreonContactgroup.class.php
```diff
@@ -222,6 +222,27 @@ public function syncWithLdap()
...
+
+                foreach ($toInsertGroups as $toInsertGroup){
+                    $this->insertLdapGroup('['.$ldaprow['ar_id'].']'.$toInsertGroup);
+                }
+
+                $res = $this->db->query("SELECT cg_id, cg_name, cg_ldap_dn FROM contactgroup
+                    WHERE cg_type = 'ldap' AND ar_id = " . $ldaprow['ar_id']);
+
+
                 while ($row = $res->fetchRow()) {
```
sc979: please use fetch() instead of fetchRow() occurrences, when working on a php7 project // 2.9.
fetchRow() is surcharged with fetch() in `centreon/class/centreonPdoStatement`

victorvassilev: thanks for the reminder, pushing that fix in a moment :)

Issue #6372, opened 2018-06-18
[clapi] export is too long when lot of parentship


kduret:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.23

**Steps to reproduce the issue:**
1. Create lot of hosts
2. Create lot of parentship between these hosts
3. Export hosts with clapi

**Describe the results you received:**
Export take too much time and CPU is 100%

**Describe the results you expected:**
Export is faster

chgautier:
Fix OK on PR #6373. Adding numerous hosts and giving each of them parent/child relations does not hinder the export execution time.

PR #6373, opened 2018-06-18
fix(clapi): avoid to order parentship several times

kduret:
Refs: #6372


chgautier:
Fix OK on branch #6372-export-host-parentship. Export with the ./centreon -u admin -p centreon -e > /tmp/clapi-export.txt command is executed rapidly.

Issue #6039, opened 2018-02-03
[trap] - no match if service with trap definition is used for multiple hosts


joschi99:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.17

**Centreon Engine version**: 1.8.1

**Centreon Broker version**: 3.0.12

**OS**: 

**Additional environment details (AWS, VirtualBox, physical, etc.):**
VMWare

**Steps to reproduce the issue:**
1. create a new passive service template with some trap events
2. create a new passive service using the service template from step 1
3. attach the service to more then one hosts

**Describe the results you received:**
the trap is no elaborated correcty: no service matching trap found (sqlite)

**Describe the results you expected:**
trap matched

**Additional information you think important (e.g. issue happens only occasionally):**
if I use the same service only for 1 host, it is working

lpinsivy:
Hi,

You are talking about SQLite database, so this problem appears on distant poller?

lpinsivy:
Can you try this fix?

```
--- /usr/share/centreon/bin/generateSqlLite.origin	2018-02-03 14:44:25.450775054 +0100
+++ /usr/share/centreon/bin/generateSqlLite	2018-02-03 14:51:16.087316220 +0100
@@ -396,26 +396,31 @@
             // Insert direct services
             $insertedServices = array();
             foreach ($result_host_services as $value) {
-                if (!isset($insertedServices[$value['service_id']])) {
-                    $stmt = $dbh_sqlite->prepare("INSERT INTO service (service_id, service_description, service_template_model_stm_id) VALUES (
-                        :service_id, :service_description, :service_template_model_stm_id)");
-                    $stmt->bindParam(':service_id', $value['service_id'], PDO::PARAM_INT);
-                    $stmt->bindParam(':service_description', $value['service_description'], PDO::PARAM_STR);
-                    $stmt->bindParam(':service_template_model_stm_id', $value['service_template_model_stm_id'], PDO::PARAM_INT);
-                    $stmt->execute();
-                    $insertedServices[$value['service_id']] = true;
-
+                if (!isset($insertedServices[$value['service_id']."-".$value['host_id']])) {
+                    if (!isset($insertedServicesUniq[$value['service_id']])) {
+                        $stmt = $dbh_sqlite->prepare("INSERT INTO service (service_id, service_description, service_template_model_stm_id) VALUES (
+                            :service_id, :service_description, :service_template_model_stm_id)");
[...]

PR #6378, opened 2018-06-19
fix(trap): no match if service is linked for multiple hosts

lpinsivy:
Insert into SQLite database all services if a SNMP trap is linked

ref: #6039


cgagnaire:
Replaced by https://github.com/centreon/centreon/pull/7447.

PR #6382, opened 2018-06-20
fix(reporting): avoid bug on partitioned tables

kduret:


chgautier:
Fix OK on branch fix-reporting-dashboard-query. No error on partitioned tables when using mysql commands.

PR #6384, opened 2018-06-20
feat(searching): persist search term for host and host group

v-radev:
Add persistence for search terms in hosts and host groups. Replace double quotes. Replace database methods.


www/include/configuration/configObject/hostgroup/listHostGroup.php
```diff
@@ -49,12 +49,20 @@
...
  * Search
  */
-$SearchTool = null;
-$search = "";
-if (isset($_POST['searchHg']) && $_POST['searchHg']) {
+$searchFilterQuery = null;
+$mainQueryParameters = [];
+$search = '';
+
+if (isset($_POST['searchHg'])) {
```
vhr: missing the second condition of if

v-radev: If the condition is like this `isset($_POST['searchHg']) && $_POST['searchHg']` there would be no way to clear the search input, because if you submit an empty string it would get the old value and it will always do search.

vhr: OK

PR #6436, opened 2018-07-04
fix(acl): Host number calculation with ACL is not correct in HG summary

lpinsivy:
For users with ACL, the SQL request get the list of all host / service couple and not only host list.


thiuyendang:
In 2.8.25 + this fix, limited user (with ACL resources configured) see no host and no host groups (could be due to ACL issue and not to the fix)

thiuyendang:
In 2.8.25, limited user can see HG summary view but number of hosts is wrong. I have generated build on PR directly and when using the container,  I've got an issue with limited user who cannot see no hosts  neither host groups at all so I don't think that it is due to the fix committed in this PR 6436. But as I don"t see hosts, I cannot check the hosts number in HG summary view.

PR #6444, opened 2018-07-05
fix(php) fix php notice undefined offset

loiclau:
awie import STPL with macro


loiclau:
fix on 91ac1eac0e9a19af7746525b31af8cea81b53992

PR #6451, opened 2018-07-09
feat(perf): Optimize acl filter calculation

sc979:
skip ACL filter calculation, when host and hostTemplate filters are already calculated


cron/centAcl-Func.php
```diff
@@ -84,53 +84,63 @@ function getFilteredPollers($host, $acl_group_id, $res_id)
...
-    global $pearDB, $hostTemplateCache;
+    global $pearDB, $hostTemplateCache, $resourceCache;
 
-    $request = "SELECT host_host_id " .
-            "FROM acl_resources_hc_relations, acl_res_group_relations, acl_resources, hostcategories_relation " .
-            "WHERE acl_resources_hc_relations.acl_res_id = acl_res_group_relations.acl_res_id " .
-            "AND acl_res_group_relations.acl_group_id = '" . $acl_group_id . "' " .
-            "AND acl_resources_hc_relations.acl_res_id = acl_resources.acl_res_id " .
+    //skipping if this $res_id was already calculated
+    if (!isset($resourceCache[$res_id])) {
```
jdelpierre: no test to know if the variable is well defined $res_id ? (no default value possible ?)

sc979: the variable is the result of a query (line 467 / centAcl.php):
`while ($res2 = $DBRESULT2->fetchRow()) {...`
but I'll add a check in centAcl.php

kduret:
Refs: #6461

PR #6465, opened 2018-07-14
Fix untracked error when using centreon-central in a docker environment

kbeaugrand:
Hi,

I was experiencing an issue when trying to create some image for the current version of Centreon web in a dockerized environment.

My issue was simple. The www/install wasn't moved to ../installDir/ bot nothing told me that.
In order to discover it, I was need to replace the @rename by a rename syntaxe.
Then I found the error was an "Invalid cross-device link" error.
This seems to be a php error on a dockerized environment. However i've replaced the line my a system call to move the install directory at the end of the step8.php file.

With those changes I was able to get my dockerized environment running well.

Furhtermore I did the same for the upgrade part.

Regards,

### DIFF
--- a/www/Themes/Centreon-2/style.css
+++ b/www/Themes/Centreon-2/style.css
@@ -609,7 +609,11 @@ input[type="submit"] {
     top: 45px;
     right: 10px;
 }
-
+#centreon-warning{
+    background: orange;
+    line-height: 30px;
+    padding-left: 30px;
+}
 form#AjaxBankBasic {
     padding-bottom: 0px;
 }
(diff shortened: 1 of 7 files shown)


ganoze:
Fine with me for the directory removal. @kduret do you have anything to say about this PR ?

kbeaugrand:
Hi, 

I come back with information about my changes. For now, I was using rmdir instead of system call. 
For the deletion detection, I was adding a step on logged part to display a warning message at the destination to the administrator to remove manually the install directory. 

Obviously, I these changes are part of the master branch differently, I guess that this PR is not relevant. 

Regards,

Issue #6461, opened 2018-07-13
ACLs calculation is too slow with lot acl resources


kduret:
---------------------------------------------------
BUG REPORT INFORMATION
---------------------------------------------------

**Centreon Web version**: 2.8.25

**Steps to reproduce the issue:**
1. Create lot of ACL resources
2. Link ACL resources with multiple ACL groups
3. Execute acl calculation script : 
     - mysql centreon_storage -e "update acl_resources SET changed = '1';"
     - /usr/bin/php -q /usr/share/centreon/cron/centAcl.php

**Describe the results you received:**
Acl calculation is too long

**Describe the results you expected:**
Acl calculation is fastest

chgautier:
Verified on PR #6495. Applied the different steps in this issue on different versions. Fix OK.

PR #6495, opened 2018-07-26
enh(perf): Optimize acl filter calculation

kduret:
Ref: #6461


chgautier:
Validation OK on branch optimize-acl-calculation. Compared with  centreon-web-2.8.25-1532444605.da32cf716 .

PR #6513, opened 2018-08-01
fix(Status Details) display true inheritance relation

loiclau:
#6482


cgagnaire:
@lpinsivy @loiclau I did a PR a long time ago, and I think its handling more cases : https://github.com/centreon/centreon/pull/5802.
What do you think ?

PR #6541, opened 2018-08-16
fix(doc): correct link references in IMP chapter

lpinsivy:


thiuyendang:
Cannot click on the links and link is not displayed as expected

thiuyendang:
Still not OK

thiuyendang:
Please, remove the "_" sign at the end of your ref link. AMO has tested without and the link is displayed properly and can be clicked.

lpinsivy:
done!

PR #6544, opened 2018-08-20
Master webservice for statistics

loiclau:


lpinsivy:
It will be better to enable/disable stats from UI!

cron/centreon-send-stats.php
```diff
@@ -0,0 +1,45 @@
+<?php
+require_once dirname(__FILE__) . '/../bootstrap.php';
```
kduret: could use __DIR__

Issue #6529, opened 2018-08-10
Services by Hostgroup wrongly sorted


CPbN:
Hi,

Monitoring page "Services by Hostgroup" is wrongly sorted.
Hostgroups are correctly sorted, but hosts inside them are not.
They are displayed based on their creation date rather than on their name.

Centreon version : 2.8.24

Thank you :+1:

lpinsivy:
Hi @CPbN can you try this:

```
--- a/usr/share/centreon/www/include/monitoring/status/ServicesHostGroups/xml/serviceGridByHGXML.php	2018-08-21 11:32:49.273171638 +0200
+++ b/usr/share/centreon/www/include/monitoring/status/ServicesHostGroups/xml/serviceGridByHGXML.php	2018-08-21 11:33:08.189170510 +0200
@@ -125,7 +125,7 @@
     $rq1 .= " AND hg.hostgroup_id IN (" . $hostgroups . ")";
 }
 $rq1 .= " AND h.enabled = 1 ";
-$rq1 .= " ORDER BY $sort_type, hg.name $order ";
+$rq1 .= " ORDER BY $sort_type, hg.name $order, host_name ASC ";
 $rq1 .= " LIMIT ".($num * $limit).",".$limit;

 $tabH = array();
```

CPbN:
Hi @lpinsivy, it solves the issue !
Thank you :+1:

thiuyendang:
cf PR 6547

PR #6547, opened 2018-08-21
fix(monitoring): Sort hosts by name ASC in serviceGridByHGXML

lpinsivy:
Ref: #6529


adr-mo:
Ok I'm on it

thiuyendang:
Reproduced in 2.8.25, fixed in 2.8.26

Issue #6528, opened 2018-08-10
Some "default page" settings do not work


CPbN:
Hi,

Some _Default page_ settings in user preferences do not work.

For example, the _Monitoring > Status Details > Services by Hostgroup_ does not.
It should redirect the user to `main.php?p=20209&o=svcOVHG` at login,
but user is redirected to `main.php?p=20209`.

I wonder if it is because in `contact` SQL table, column `default_page` is an `int` rather than a `varchar`, thus can't store `20209&o=svcOVHG` but only `20209`.

Centreon version : 2.8.24

Thank you 👍

lpinsivy:
Yes when you connect it must be add Hi @CPbN can you try:

```
--- a/usr/share/centreon/www/class/centreonAuth.class.php
	2018-08-21 11:53:40.798097041 +0200
+++ b/usr/share/centreon/www/class/centreonAuth.class.php	2018-08-21 13:17:40.984796623 +0200
@@ -280,6 +280,11 @@
         }
         if ($DBRESULT->numRows()) {
             $this->userInfos = $DBRESULT->fetchRow();
+            $DBRESULT2 = $this->pearDB->query("SELECT topology_url_opt FROM topology WHERE topology_page = " . $this->userInfos["default_page"]);
+            if ($DBRESULT2->numRows()) {
+                $data = $DBRESULT2->fetchRow();
+                $this->userInfos["default_page"] .= $data["topology_url_opt"];
+            }

             if ($this->userInfos["contact_oreon"]
                 || ($this->userInfos["contact_oreon"] == 0 && $this->source == 'API')) {
@@ -325,6 +330,11 @@
             );
             if ($DBRESULT->numRows()) {
                 $this->userInfos = $DBRESULT->fetchRow();
+                $DBRESULT2 = $this->pearDB->query("SELECT topology_url_opt FROM topology WHERE topology_page = " . $this->userInfos["default_page"]);
+                if ($DBRESULT2->numRows()) {
+			$data = $DBRESULT2->fetchRow();
+                        $this->userInfos["default_page"] .= $data["topology_url_opt"];
+		}
             }
         } else {
             if ($this->debug) {
```

lpinsivy:
Yes the default page is only loaded on authentication of the user.

Can you create another issue to add this feature?

CPbN:
Yes sure -> #6548
Thx !

PR #6551, opened 2018-08-21
fix(topology): Add topology url option when loading default page

lpinsivy:
* Get from topology table the topology_url_opt field to add in default url page the options

Ref: #6528


adr-mo:
Starting validation

PR #6586, opened 2018-09-01
change the bash interpreter for the native sh -- references to GPLv3

lgcosta:
This PR is intended to remove legacy GPLv3 license references from the `bash` interpreter. Although much more complete and with more options, there is the license bait.

Using native `sh` also gives us better support for other systems, such as BSD (FreeBSD, OpenBSD, etc.. ) that do not use `bash` in their native shell (exactly on about of the license), and we eliminate another dependency.


ganoze:
Closed in favor of #7911 . This PR will not make it to releases earlier to 19.10 and it was to cumbersome to cherry-pick this commit from 2.8.x to master.

PR #6593, opened 2018-09-06
Remove spaces at end of line

lgcosta:
Removes unneeded spaces at end of line and helps maintain style guide.


sc979:
this PR, may cause many conflicts

lgcosta:
Yes, maybe wait for some other PR's and then i can fix the conflicts here in this PR.

kduret:
sorry but we can't merge this PR cause this will break the git history.
We prefer to fix coding style when we work directly on each files.

PR #6620, opened 2018-09-19
Update Nagios Plugins path

ganoze:
As Nagios Plugins are now placed under /usr/lib64 on CentOS 7, their path have to be updated if necessary.

### DIFF
--- a/www/install/php/Update-2.8.26_to_18.10.0.php
+++ b/www/install/php/Update-2.8.26_to_18.10.0.php
@@ -1,0 +1,54 @@
+<?php
+/*
+ * Copyright 2005-2018 Centreon
+ * Centreon is developed by : Julien Mathis and Romain Le Merlus under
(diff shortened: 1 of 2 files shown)


www/install/php/Update-2.8.26_to_18.10.0.php
```diff
@@ -0,0 +1,54 @@
...
+ * For more information : contact@centreon.com
+ *
+ *
+ */
+
+// Retrieve current Nagios plugins path.
+$query = "SELECT value FROM options WHERE `key`='nagios_path_plugins'";
+$result = $pearDB->query($query);
+$row = $result->fetchRow();
+error_log(print_r($row, true));
```
kduret: remove debug :)

ganoze: Good catch !

PR #6638, opened 2018-09-21
apply intervals on header refresh

victorvassilev:


www/front_src/src/components/header/index.js
```diff
@@ -57,35 +57,42 @@ class TopHeader extends Component {
     });
   };
 
+  _updateMenuData = () => {
```
kduret: should be rename to _updateHeaderData to avoid confusion with the left side menu

PR #6653, opened 2018-09-26
Mon-datepicker-split-method

sc979:
I Need to split the method for MBI, to avoid re-initializing the datepickers when user's format need to be set again.

### DIFF
--- a/www/include/common/javascript/datepicker/localizedDatepicker.js
+++ b/www/include/common/javascript/datepicker/localizedDatepicker.js
@@ -33,14 +33,13 @@
 */
 
 /**
- * Used for initializing datepicker with an alternative format field
+ * Used to initialize datepicker with an alternative format field
  *
  * @param className string : tag class name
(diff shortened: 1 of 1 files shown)


www/include/common/javascript/datepicker/localizedDatepicker.js
```diff
@@ -53,7 +52,30 @@ function initDatepicker(className, altFormat, defaultDate) {
...
     }
 
-    /* Getting user's localization and loading corresponding library */
+    setUserFormat();
+
+    // initializing datepicker and the hidden alternate format field
+    jQuery("." + className).each(function () {
+        // finding the alternative field
+        var altName = jQuery(this).attr("name");
+        var alternativeField = "input[name=alternativeDate" + altName[0].toUpperCase() + altName.slice(1) + "]";
```
jdelpierre: if altName, the table doesn't have an index 0 it won't crash?

PR #6658, opened 2018-09-26
Enhancement about menu

callapa:


callapa:
Replace insert query by update query to avoid conflict during upgrade from older version

PR #6662, opened 2018-09-26
MON-2858

danielhalupka:


callapa:
You can use this command to check the coding style :
`vendor/bin/phpcs -p --standard='PSR2' www/include/configuration/configServers/listServers.php`

PR #6665, opened 2018-09-26
fix(menu): Fix the menu order

callapa:
Fix the menu order to be the same than in 2.8.x version


src/CentreonLegacy/Core/Menu/Menu.php
```diff
@@ -166,15 +172,25 @@ public function getMenu()
                     'is_react' => $row['is_react']
                 ];
                 if (!is_null($row['topology_group']) && isset($groups[$levelTwo][$row['topology_group']])) {
+                    /**
```
sc979: no need to comment each case. Put your comment before the if ;)

PR #6675, opened 2018-09-27
fix(front): fix datepicker init

kduret:


sc979:
Defaultdate should ne 0 not O

PR #6689, opened 2018-10-01
fix(admin): reload parent window when the module is updated or installed

vhr:
Resolve MON-2877

### DIFF
--- a/www/include/options/oreon/modules/installForm.php
+++ b/www/include/options/oreon/modules/installForm.php
@@ -113,23 +113,43 @@ if (file_exists($moduleInfoObj->getModulePath() . '/infos/infos.txt')) {
 $form1 = new HTML_QuickFormCustom('Form', 'post', "?p={$p}");
 
 if ($form1->validate()) {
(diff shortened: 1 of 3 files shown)


www/include/options/oreon/modules/installForm.php
```diff
@@ -113,25 +113,45 @@
...
+
+    $tpl->display('reload.ihtml');
+    return;
 } elseif ($o == 'i' && !$moduleInfoObj->getInstalledInformation($name)) {
     $form1->addElement('submit', 'install', _('Install Module'), array('class' => 'btc bt_success'));
     $redirect = $form1->addElement('hidden', 'o');
     $redirect->setValue('i');
 }
 
+if (array_key_exists('msg', $_GET) && $_GET['msg']) {
```
sc979: isn't it a $_POST instead a $_GET ?

vhr: nope this parameter is in URL query after the POST request the page will be redirected to new URL with GET request

PR #6700, opened 2018-10-01
Test doubleclick

EdytaSki:


www/front_src/src/components/navigation/index.js
```diff
@@ -8,14 +8,14 @@ import axios from "../../axios";
...
-  navService = axios("internal.php?object=centreon_menu&action=menu");
 
-  state = {
-    active: false,
-    initiallyCollapsed: false,
-    selectedMenu: {},
-    menuItems: []
-  };
+  navService = axios("internal.php?object=centreon_menu&action=menu");
+    state = {
```
kduret: bad indentation

www/front_src/src/components/navigation/index.js
```diff
@@ -43,6 +43,29 @@ class NavigationComponent extends Component {
     });
   };
 
+  handleDoubleClickItem = index => {
+    switch (index) {
+      case 0:
+        window.location.href='main.php?p=103';
```
kduret: please use gotopage method to avoid a full reload of the page

PR #6720, opened 2018-10-03
Mon 2873 fix blank page

callapa:


callapa:
Fix error on topology_opt that have been replace by topology_url_opt.

PR #6721, opened 2018-10-03
Multistepform finalize

v-radev:


lib/perl/centreon/script/centcore.pm
```diff
@@ -336,8 +336,7 @@ sub startWorker($) {
     my $username = $data->{'contact_alias'};
     my $passwordEnc = $data->{'contact_passwd'};
 
-    my $cmdexec = "php $self->{centreonDir}/bin/centreon -u $username -p $passwordEnc -s -o CentreonWorker -a processQueue >> /var/log/centreon/worker.log";
-
+    my $cmdexec = "php72 $self->{centreonDir}/bin/centreon -u $username -p $passwordEnc -w -o CentreonWorker -a processQueue >> /var/log/centreon/worker.log";
```
vhr: why?

kduret: bin/centreon must ahve the right shebang. So don't use explicitly your php binary

www/front_src/src/route-components/pollerStep2/index.js
```diff
@@ -69,7 +69,7 @@ class PollerStepTwoRoute extends Component {
         <ProgressBar links={links} />
         <Form
           pollers={pollers}
-          initialValues={pollerData}
+          initialValues={{...pollerData}}
```
kduret: let pollerData

src/CentreonRemote/Application/Webservice/CentreonTaskService.php
```diff
@@ -179,14 +189,6 @@ public function postAddImportTaskWithParent()
...
             if (!$parent_id) {
                 throw new Exception('Missing parent_id parameter');
-            } elseif (!$params) {
-                throw new Exception('Missing params parameter');
-            } elseif (!array_key_exists('pollers', $params)) {
-                throw new Exception('Missing pollers list in params');
-            } elseif (!array_key_exists('pollers', $params)) {
-                throw new Exception('Missing pollers list in params');
-            } elseif (!array_key_exists('server', $params)) {
-                throw new Exception('Missing server ID in params');
```
vhr: Do we have validation of $params in the `addTask` method?

victorvassilev: no, it just needs to be an array, can be an empty one

PR #6723, opened 2018-10-03
fix(autologin): display autologin only if defined

EdytaSki:


www/front_src/src/components/userMenu/index.js
```diff
@@ -74,15 +97,11 @@ class UserMenu extends Component {
                   </span>
                 </li>
                 <React.Fragment>
```
kduret: don't use react fragment for this usage

www/front_src/src/components/userMenu/index.js
```diff
@@ -74,15 +97,11 @@ class UserMenu extends Component {
...
-                  <button class="submenu-user-button" onClick={this.onCopy.bind(this)}>Copy autologin link <span className={`iconmoon btn-logout-icon ${(copied ? 'icon-copied' : 'icon-copy')}`}></span></button>
-                  <textarea id="autologin-input" style={
-                    {
-                      width: 0,
-                      height: 0,
-                      position: 'fixed',
-                      top: -100
-                    }
-                  }>{value}</textarea>
+                  {copyToClipboardBtn}
```
kduret: don't create the component if autologin is not defined

www/front_src/src/helpers/autoLoginTokenGenerator.js
```diff
@@ -0,0 +1,49 @@
+import React from "react";
```
kduret: don't import react

PR #6743, opened 2018-10-05
fix(api): fix getter of poller states (database and latency)

kduret:


www/api/class/centreon_topcounter.class.php
```diff
@@ -738,6 +738,7 @@ protected function pollersStatusList()
                 $listPoller[$row['instance_id']]['latency']['state'] = 2;
                 $listPoller[$row['instance_id']]['database']['time'] = $row['stat_value'];
             } elseif ($row['stat_value'] >= 60) {
+                $listPoller[$row['instance_id']]['latency']['state'] = 1;
```
sc979: `$listPoller[$row['instance_id']]['latency']['state']` was already set on the next line

PR #6747, opened 2018-10-07
new(translation): Bakoffice to get all translations extracted on all messages.po

callapa:
MON-2903


kduret:
could you wait front-end translation before merge this PR please

PR #6760, opened 2018-10-08
Fix/menu font size

EdytaSki:
Making font size bigger because we can not read it


kduret:
rename your PR ;) 
Like "enh(ui): increase font size of menu entry"

PR #6761, opened 2018-10-08
fix(menu): Fix the ACLs of menu

callapa:
Avoid the first menu level having more privileges than the first child

### DIFF
--- a/www/class/centreonACL.class.php
+++ b/www/class/centreonACL.class.php
@@ -443,41 +443,81 @@ class CentreonACL
     }
     
     /**
-     * Use to check and fix if in the topology, a parent has access rights that 
-     * can be higher than children when they have the same URL.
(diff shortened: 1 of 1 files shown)


www/class/centreonACL.class.php
```diff
@@ -439,6 +439,95 @@ private function setTopology()
...
+     * Use to check and fix if in the topology, a parent has access rights that
+     * can be higher than children when they have the same endpoint.
+     */
+    private function checkTopology(): void
+    {
+        if (!empty($this->topology)) {
+            /**
+             * Filter to keep the first child available per level.
+             */
+            $getFirstChildPerLvl = function (array $topologies): ?array {
```
kduret: It seems that there is no case where you return "null".
So return type should be "array" instead of "?array"

PR #6763, opened 2018-10-09
feat(acl): add webservice to get react pages filtered with acl

victorvassilev:


src/Centreon/Domain/Repository/TopologyRepository.php
```diff
@@ -29,4 +29,79 @@ public function enableMenus(): bool
...
+                        while ($topo_page = $DBRESULT3->fetchRow()) {
+                            $topologyUrls[] = $topo_page["topology_url"];
+                        }
+                        $DBRESULT3->closeCursor();
+                    }
+                }
+            }
+        }
+
+        return $topologyUrls ?: null;
```
kduret: why not return an empty array ?
I think it's easier to manage in the frontend part

victorvassilev: you are right! will update

PR #6774, opened 2018-10-09
fix(user profile): Fix the default page on login

callapa:
Correction of the user profile form to take into account the default page after login.


www/include/Administration/myAccount/formMyAccount.php
```diff
@@ -129,44 +129,93 @@
...
+};
+
+$topologies = $createTopologyTree($acls);
+
+/**
+ * Retrieve the name of all topologies available for this user
+ */
+$query = "SELECT topology_page, topology_name "
+    . "FROM topology "
+    . "WHERE topology_page IN ($aclUser)";
```
victorvassilev: can we add AND is_react = "0" ? Because we are going to be adding the react pages in topology as well.

PR #6779, opened 2018-10-10
feat(translation): add chinese translation

lpinsivy:


lpinsivy:
@ganoze help.po will be provided latter.

@sc979 I suppose yes. Can you add it in this PR?

PR #6786, opened 2018-10-11
fix(ui): retrieve the same page on browser refresh

victorvassilev:


www/front_src/src/route-components/module/index.js
```diff
@@ -32,6 +32,11 @@ class ModuleRoute extends Component {
     }
   }
 
+  handleHref = (event) => {
+    let href = event.detail.href;
+    window.history.pushState(null,null, href);
```
sc979: please add a space after the first "null," param

PR #6788, opened 2018-10-11
MON-2949 Cannot close submenu once clicked

kovacmarko94:


www/front_src/src/components/navigation/index.js
```diff
@@ -67,9 +67,9 @@ class NavigationComponent extends Component {
         let { menuItems } = this.state;
 
         Object.keys(menuItems).forEach(key => {
-          menuItems[key].toggled = false;
-        })
-        menuItems[index].toggled = !menuItems[index].toggled;
+          menuItems[key].toggled = key === index ?
+            !menuItems[index].toggled : menuItems[key].toggled = false;
```
kduret: why reassign value to menuItems[key].toggled ?
==> !menuItems[index].toggled :  false;

PR #6791, opened 2018-10-11
MON-2953 First Administration => Parameters submenu is main

kovacmarko94:


kduret:
I guess it breaks unit tests
Please update these in tests/php/CentreonLegacy/Core/Menu/MenuTest.php

ganoze:
Isn't this commit based on an invalid bug report ? In my case the first submenu in blue in called "main". Shouldn't it have a better name ?

first_submenu_main

PR #6809, opened 2018-10-14
feat(database): add support for MariaDB SQL_MODE with STRICT_TRANS_TABLES

jorgemorgado:
Since MariaDB 10.2 the default SQL_MODE has STRICT_TRANS_TABLES enabled. This means that statements with invalid or missing data are aborted and rolled back. For more information see https://mariadb.com/kb/en/library/sql-mode/. This commit fixes those queries (mostly INSERT INTO queries but there is also one case where NULL wrongly quoted). Additionally, since my editor runs a PHP lint, the changed files got trailing spaces removed.

### DIFF
--- a/bin/centreon-translations.php
+++ b/bin/centreon-translations.php
@@ -1,0 +1,151 @@
+<?php
+/**
+ * Copyright 2005-2018 Centreon
+ * Centreon is developped by : Julien Mathis and Romain Le Merlus under
+ * GPL Licence 2.0.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation ; either version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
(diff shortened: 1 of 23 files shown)


jorgemorgado:
I found one more INSERT INTO query that needed to be fixed but meanwhile had to rebase. There are still more queries to review (e.g., UPDATE queries) but the proposed changes will not break the code. It's either the same behaviour as before (if STRICT_TRANS_TABLES is disabled), or the correct behaviour (if enabled).

bouda1:
Thank you for all this work. Even if we apply your patches, we won't be able to switch the database in strict mode because centreon-broker, for example, needs also it to be not strict.
I hope we will be able to make that switch after the new Centreon version will be released.
Thank you again ! Best regards.

jorgemorgado:
Yes that's fine. I just wanted to know if you are okay with the current compromise solution I'm taking - the "ideal" INSERT INTO queries should explicitly specify the fields to be inserted and skip the primary key. Or even better, use prepared statements. But my approach was to modify as less code as possible and let you decide on the best way to proceed.
In any case, I will see if I can identify other places where strict mode will cause a query to fail. Thanks for the feedback.

lpinsivy:
Replaced by #7544

PR #6819, opened 2018-10-15
fix(front): redirect properly to login page if session is expired

kduret:


www/class/centreonSession.class.php
```diff
@@ -109,17 +109,25 @@ public static function checkSession($sessionId, CentreonDB $pearDB)
...
-            . 'WHERE `session_id` = "' . session_id() . '" ';
-        $pearDB->query($query);
+        $sessionId = session_id();
+
+        if ($this->checkSession($sessionId, $pearDB) === 1) {
+            /* Update last_reload parameter */
+            $query = 'UPDATE `session` '
+                . 'SET `last_reload` = "' . time() . '", '
+                . '`ip_address` = "' . $_SERVER["REMOTE_ADDR"] . '" '
+                . 'WHERE `session_id` = "' . session_id() . '" ';
```
jdelpierre: It's better to use your declaration $sessionId :)

PR #6834, opened 2018-10-17
fix(translation): Add translation for button

callapa:
Add translation for button with label 'Add server with wizard' and create a php file with all React translations to use with messages.pot


www/install/react_translation.php
```diff
@@ -0,0 +1,77 @@
+<?php
+/*
+ * Copyright 2005-2018 Centreon
+ * Centreon is developped by : Julien Mathis and Romain Le Merlus under
```
sc979:
```suggestion
 * Centreon is developed by : Julien Mathis and Romain Le Merlus under
```

PR #6836, opened 2018-10-17
fix(remote-server) Minor fixes for remote server worker

lpinsivy:
* Prevent empty centreon_path for Poller Display migration because poller configuration is present and the user don't use the wizard (centreon_path => NULL in database)
* Add more debug information

### DIFF
--- a/src/CentreonRemote/Application/Clapi/CentreonWorker.php
+++ b/src/CentreonRemote/Application/Clapi/CentreonWorker.php
@@ -118,7 +118,7 @@ class CentreonWorker implements CentreonClapiServiceInterface
          */
         $params = unserialize($task->getParams())['params'];
         $centreonPath = trim($params['centreon_path'], '/');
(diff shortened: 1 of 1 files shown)


src/CentreonRemote/Application/Clapi/CentreonWorker.php
```diff
@@ -118,14 +118,16 @@ public function createRemoteTask(int $taskId)
          */
         $params = unserialize($task->getParams())['params'];
         $centreonPath = trim($params['centreon_path'], '/');
+        $centreonPath = (isset($centreonPath) && strcmp($centreonPath, '') != 0) ? $centreonPath : '/centreon';
```
vhr: an isset function is not necessary because on row 120 you are defined the variable $centreonPath
```suggestion
        $centreonPath = $centreonPath ? $centreonPath : '/centreon';
```

PR #6861, opened 2018-10-18
fix(configuration): duplication of host and hostgroup

victorvassilev:
Force reload of acl before listing of data on page

### DIFF
--- a/www/include/configuration/configObject/host/host.php
+++ b/www/include/configuration/configObject/host/host.php
@@ -117,7 +117,6 @@ switch ($o) {
     case "m":
         multipleHostInDB(isset($select) ? $select : array(), $dupNbr);
         $acl = $centreon->user->access;
-        $aclDbName = $acl->getNameDBAcl('broker');
         $hgs = $acl->getHostGroupAclConf(null, 'broker');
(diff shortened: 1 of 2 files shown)


www/include/configuration/configObject/hostgroup/hostGroup.php
```diff
@@ -106,6 +106,11 @@ function mywrap($el)
         break;
     case "m":
         multipleHostGroupInDB(isset($select) ? $select : array(), $dupNbr);
+        $acl = $centreon->user->access;
+        $aclDbName = $acl->getNameDBAcl();
```
kduret: this is used ?

www/include/configuration/configObject/host/host.php
```diff
@@ -116,6 +116,11 @@
         break;
     case "m":
         multipleHostInDB(isset($select) ? $select : array(), $dupNbr);
+        $acl = $centreon->user->access;
+        $aclDbName = $acl->getNameDBAcl('broker');
```
kduret: this is used ?

victorvassilev: acl yes, the aclDbName maybe doesnt need to be updated

PR #6865, opened 2018-10-19
fix(js): wait content load before fire jquery event

danielhalupka:
Issue was that jQuery fired before content was loaded, so we check if it's loaded and then fire it one more time to let JS do it's magic


www/Themes/Centreon-2/style.css
```diff
@@ -1414,7 +1414,7 @@ table.ListTableSmallArg td {
 .ListTable,.ListTableMedium,.ListTableSmall,.ListTableMultiTp {border-collapse:separate;border-spacing:0;border-style:solid;empty-cells:show;}
 .ListTable,.ListTableMedium,.ListTableSmall {border-width:1px;}
 .ListTable td,.ListTableMedium td,.ListTableSmall td,.ListTableSmallArg {padding-bottom:2px;padding-top:2px;}
-.ListTable td {padding: 4px; overflow: hidden; text-overflow: ellipsis;}
+.ListTable td {padding: 4px;}
```
kduret: be carefull that it does not break other listing pages

danielhalupka: This was added by Edyta and NIkola, so that's an revert

PR #6870, opened 2018-10-19
fix(remote): empty bam broker config on new remote server

vhr:
Resolve MON-3065


Pontissalien:
@kduret better chance that there is no conflit with those values than with previous 0 and 1 values of config_group_id

Pontissalien:
Right now this should fix the problem for the 18.10.
But later it will be possible that these values conflits again if main part of the config gets more outputs.

PR #6883, opened 2018-10-22
fix(DT): external_cmd timestamp fix for the widgets

sc979:
resolve: MON-3133


www/class/centreonUtils.class.php
```diff
@@ -148,7 +148,7 @@ public static function conditionBuilder($query, $condition, $or = false)
      * @return int
      * @throws Exception
      */
-    public function getDateTimeTimestamp($datetime)
+    public static function getDateTimeTimestamp($datetime)
```
sc979: method only used in host and service monitoring widgets

PR #6894, opened 2018-10-23
fix(login): inception, new header.php for iframe

victorvassilev:
not to allow iFrame to get to login.php page

### DIFF
--- a/www/include/core/header/header.iframe.php
+++ b/www/include/core/header/header.iframe.php
@@ -1,0 +1,248 @@
+<?php
+/*
+ * Copyright 2005-2015 Centreon
+ * Centreon is developped by : Julien Mathis and Romain Le Merlus under
+ * GPL Licence 2.0.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation ; either version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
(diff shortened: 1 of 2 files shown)


www/include/core/header/header.iframe.php
```diff
@@ -0,0 +1,248 @@
...
+require_once $classdir . '/centreonFeature.class.php';
+require_once SMARTY_DIR . "Smarty.class.php";
+
+/*
+ * Create DB Connection
+ *  - centreon
+ *  - centstorage
+ */
+$pearDB = new CentreonDB();
+$pearDBO = new CentreonDB("centstorage");
```
sc979:
```suggestion
$pearDBO = $dependencyInjector['realtime_db'];
```

www/include/core/header/header.iframe.php
```diff
@@ -0,0 +1,248 @@
...
+$time_start = microtime_float();
+
+$advanced_search = 0;
+
+/*
+ * Include
+ */
+include_once(realpath(dirname(__FILE__) . "/../../../../bootstrap.php"));
+
+require_once "$classdir/centreonDB.class.php";
```
sc979: when using dependencyInjector, you don't need this class anymore

sc979: ! -> $classdir may not be defined
I found only one affectation in the "centreon.config.php.template" . But I may be wrong. Why don't you use a  $path = __DIR__ . "XXX" and use it instead ?

sc979: PSR2 -> "$classdir/centreonDB.class.php" -> $classdir . "/centreonDB.php" like line 63

kduret:
I close This PR cause it's not mergeable anymore.
And I think we've may some fixes about login page redirection.
We'll reopen this if problem reappears

PR #6931, opened 2018-11-02
fix(doc): correct typo

lpinsivy:
Correct many typo in FR & EN documentation


lpinsivy:
@chgautier our changes are good but I can't validate those. Can you put "Approved"?

PR #6985, opened 2018-11-21
fix(externalcmd): Create centcore file by action

lpinsivy:
To remove concurrent access to centcore.cmd file, create a file in /var/lib/centreon/centcore by action

### DIFF
--- a/www/class/centreonExternalCommand.class.php
+++ b/www/class/centreonExternalCommand.class.php
@@ -126,7 +126,7 @@ class CentreonExternalCommand
             if ($this->debug) {
                 print "COMMAND BEFORE SEND: $str_remote";
             }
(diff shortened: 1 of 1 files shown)


lpinsivy:
Use microtime instead of time

www/class/centreonExternalCommand.class.php
```diff
@@ -126,7 +126,7 @@ public function write()
             if ($this->debug) {
                 print "COMMAND BEFORE SEND: $str_remote";
             }
-            $result = file_put_contents($varlib . '/centcore.cmd', $str_remote, FILE_APPEND);
+            $result = file_put_contents($varlib . '/centcore/' . microtime(true) . '-externalcommand.cmd', $str_remote, FILE_APPEND);
```
sc979:
```suggestion
            $result = file_put_contents($varlib . '/centcore/' . microtime(true) . '-externalcommand' . $cmd, $str_remote, FILE_APPEND);
```

PR #7004, opened 2018-11-26
fix(acl): service_id is no longer NULL

lpinsivy:
Since PR #6366, all values for host are: <host_id>,0 and not <host_id>,NULL


www/include/configuration/configObject/host/listHost.php
```diff
@@ -241,7 +241,7 @@
 if (!$centreon->user->admin) {
     $aclFrom = ", {$aclDbName}.centreon_acl acl";
     $aclCond =
-        ' AND h.host_id = acl.host_id AND acl.service_id IS NULL '
```
kduret: On my platform, service_id is NULL
I don't understand when it can be 0, cause default value of this column is NULL

PR #7043, opened 2018-12-12
fix(conf): protect hostname resolver from XSS

vhr:
Resolve MON-3228

### DIFF
--- a/www/include/configuration/configObject/host/resolveHostName.php
+++ b/www/include/configuration/configObject/host/resolveHostName.php
@@ -57,4 +57,9 @@ if (isset($sid)) {
 /**
  * Resolving host name
  */
-echo gethostbyname($_GET['hostName']);
+$hostName = str_replace(
+    array('<', '>'),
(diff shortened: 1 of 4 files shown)


www/include/configuration/configObject/host/resolveHostName.php
```diff
@@ -57,4 +57,9 @@
 /**
  * Resolving host name
  */
-echo gethostbyname($_GET['hostName']);
+$hostName = str_replace(
+    array('<', '>'),
+    array('&lt;', '&gt;'),
+    $_GET['hostName']
+);
+echo gethostbyname($hostName);
```
kduret: should not be better echo htmlentities(gethostbyname($hostName)); ?

vhr: If we do that this will be the result:
screenshot from 2018-12-12 10-56-55

I propose to strip the tags and remove all chars that are not compatible with the hostname.

sc979:
QA validationfailed :'(

PR #7046, opened 2018-12-12
Mon 3281 set default language to browser

sc979:
Resolve : MON-3281


www/install/sql/centreon/Update-DB-18.10.2_to_18.10.3.sql
```diff
@@ -2,4 +2,8 @@
 UPDATE `informations` SET `value` = '18.10.3' WHERE CONVERT( `informations`.`key` USING utf8 ) = 'version' AND CONVERT ( `informations`.`value` USING utf8 ) = '18.10.2' LIMIT 1;
 
 -- Remove old progressbar lib
-DELETE FROM topology_JS WHERE PathName_js LIKE '%aculous%';
\ No newline at end of file
+DELETE FROM topology_JS WHERE PathName_js LIKE '%aculous%';
+
+-- Changing contact_lang default value to 'browser' and each of its value from NULL to 'browser'
+ALTER TABLE contact CHANGE `contact_lang` `contact_lang` varchar(255) DEFAULT 'browser';
```
callapa: Why do you repeat the same column ?

sc979: because of :
https://dev.mysql.com/doc/refman/5.6/en/alter-table.html
ALTER TABLE [table] CHANGE [old_col_name] [new_col_name] [column_definition]

www/api/class/centreon_datepicker_i18n.class.php
```diff
@@ -0,0 +1,64 @@
...
+     *
+     * @return string|null
+     */
+    public function getDatepickerLibrary()
+    {
+        $data = $this->arguments['data'];
+        $path = __DIR__ . "/../../include/common/javascript/datepicker/";
+
+        // check if the fullLocale library exist
+        $datepickerFileToFound = "datepicker-" . substr($data, 0, 5) . ".js";
```
kduret: I think you could use preg_match to avoid double substr

PR #7047, opened 2018-12-13
Normalize names of lang directories

ganoze:
This is required to automatize the creation of new lang files without modifying the packaging scripts.


lpinsivy:
@ganoze can you rebase because you include another PR in your

PR #7059, opened 2018-12-17
automatically add/update PHP files with licence text with php-cs-fixer

vhr:


kduret:
I close this issue.
I think this kind of checks needs to be integrated in the CI

PR #7061, opened 2018-12-17
Mon 3224 sql inj virtual metric

vhr:


www/include/views/virtualMetrics/DB-Func.php
```diff
@@ -179,40 +270,68 @@ function insertVirtualMetric()
...
-    $DBRESULT = $pearDB->query($rq);
-    $DBRESULT = $pearDB->query("SELECT MAX(vmetric_id) FROM virtual_metrics");
-    $vmetric_id = $DBRESULT->fetchRow();
-    $vmetric_id = $vmetric_id["MAX(vmetric_id)"];
-    return $vmetric_id;
+    $pearDB->query($rq);
+    $DBRESULT = $pearDB->query(
+        'SELECT MAX(vmetric_id) AS vmetric_id FROM virtual_metrics'
+    );
+    return (int) $DBRESULT->fetchRow()['vmetric_id'];
```
kduret: not compatible with php 5.3 (centos 6)

vhr: fixed

PR #7072, opened 2018-12-19
MON-3298 enhance utf-8 check

Pontissalien:
UTF-8 encoding control use lot of memory and engine conf generation fails when there are few thousands of resources
we'd better use the default php implementation instead (mb_detect_encoding)


thiuyendang:
Validated with rpms 1548423274.e6aff13a0.el7.
Configured 122 hosts and 1707 services including some with special characters
Before fix : error reproduced
After fix, no more error only with default value of the memory limit in php.ini file

PR #7079, opened 2018-12-21
fix(status details): Fix hide password in command line

lpinsivy:
Previous method: get the value of the custom macro password type and replace all matches in command line

New method: Get the option name in the command line and replace only option/value couple 

Tested on different command lines but need several tests before to be merged


www/include/monitoring/objectDetails/common-func.php
```diff
@@ -1,6 +1,6 @@
 <?php
-/*
- * Copyright 2005-2016 Centreon
+/**
+ * Copyright 2005-2018 Centreon
  * Centreon is developped by : Julien Mathis and Romain Le Merlus under
```
sc979:
```suggestion
  * Centreon is developed by : Julien Mathis and Romain Le Merlus under
```

www/include/monitoring/objectDetails/common-func.php
```diff
@@ -45,31 +54,40 @@ function hidePasswordInCommand($command_name, $host_id, $service_id)
 
     $pearDBStorage = new CentreonDB('centstorage');
 
-    /* Get executed command lines */
+    /* Get command line with macro */
+    $query_command_line = "SELECT command_line FROM command WHERE command_name = '" .
+    $pearDB->escape($command_name) . "'";
```
sc979: binding the values prevent from SQL injection

lpinsivy: Yes but all parameters are from PHP and not user ;)

PR #7083, opened 2018-12-21
fix(secu): Authenticated Remote Code Execution in getStats.php

gquere:
Due to the way preg_replace handles backslahes, the expression:

    function escape_command($command)
    {
        return preg_replace("/[\\\$|`]/", "", $command);
    }

wouldn't work as expected because 3 backslashes are needed to properly refer to the backslash inside the regex.

Therefore the following GET parameter would lead to arbitrary remote code execution:
http://x.x.x.x/centreon/include/Administration/corePerformance/getStats.php?key=active_service_check&ns_id=1=/var/lib/centreon/nagios-perf/perfmon-1/nagios_active_service_execution.rrd:Min:AVERAGE \%26 touch /tmp/das \%26

Because it would be interpreted as \&#039; which still makes it valid for a shell to execute.

I'm not very confident in the multiple layers of filtering in this PHP file. It really needs to NOT use GET parameters.

### DIFF
--- a/www/include/Administration/corePerformance/getStats.php
+++ b/www/include/Administration/corePerformance/getStats.php
@@ -79,7 +79,7 @@ foreach ($_GET as $key => $get) {
 
 function escape_command($command)
 {
-    return preg_replace("/[\\\$|`]/", "", $command);
+    return preg_replace("/[\\\\$|`]/", "", $command);
 }
 
     require_once realpath(dirname(__FILE__) . "/../../../../config/centreon.config.php");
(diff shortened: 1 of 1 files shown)


adr-mo:
Thanks for your contribution @gquere 
This file is not used anymore and for a while.

This PR https://github.com/centreon/centreon/pull/7195 will remove the file for the **18.10.x** and **master**. An other PR https://github.com/centreon/centreon/pull/7271 for the **2.8.x**

adr-mo:
Thanks for your contribution @gquere 
But the file getStats.php is no longer since graphs were migrated to c3.js.
This is therefore dead code that needs to be removed from branches:
- 2.8.x
- 18.10.x
- master

This PR https://github.com/centreon/centreon/pull/7195 is made for the 18.10.x and master branch
This PR is made for the 2.8.x branch https://github.com/centreon/centreon/pull/7271

Regards

PR #7084, opened 2018-12-21
fix(secu): Type juggling can lead to authentication bypass in (very) …

gquere:
…rare cases

The '==' operator is a loose comparison operator, whereas the '===' does a strict comparison.
The truth table of the '==' operator is quite ugly (https://www.owasp.org/images/6/6b/PHPMagicTricks-TypeJuggling.pdf) and can lead to undesirable integer promotion results.

For instance, when loose comparing something that looks like integers, PHP will promote both strings to integers.

Example: if the user's token starts with "0e" and is followed by numbers, it is bypassable by the string "00" because in PHP, ("0e123456" == "00") is TRUE.
    http://192.168.56.3/centreon/index.php?p=1&autologin=1&useralias=admin&token=00

Altough, the probabilty this can happen to e.g. the token is very low:
    1/62 * (2/62) * (10/62)^6

### DIFF
--- a/www/class/centreonAuth.class.php
+++ b/www/class/centreonAuth.class.php
@@ -154,7 +154,7 @@ class CentreonAuth
      */
     protected function checkPassword($password, $token = "", $autoimport = false)
     {
-        if ((strlen($password) == 0 || $password == "") && $token == "") {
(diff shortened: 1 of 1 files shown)


adr-mo:
New branch created for the validation **MON-3306-gquere-authentication_bypass**

PR #7086, opened 2018-12-21
fix(secu): Multiple authenticated SQL injections in img_gantt.php

gquere:
Unescaped direct use of $_GET parameters inside a non prepared statement allowing stacked queries.

Recommend usage of a prepared execute() rather than using query().


loiclau:
#7200 thanks for the participation

PR #7087, opened 2018-12-21
fix(secu): Authenticated SQL injection in makeXML_ListServices.php

gquere:
Unescaped direct use of a $_GET parameter inside a non prepared statement allowing stacked queries.

For instance, exploitable (when authenticated) with this request:
http://192.168.56.3/centreon/include/views/graphs/common/makeXML_ListServices.php?host_id=1'; DO SLEEP(5); -- -


sc979:
Hi gquere,

This file is unused since a previous major release. And needs to be deleted from the master's and 18.10.x's branches instead of being fixed.
Please check this PR : https://github.com/centreon/centreon/pull/7229

Best Regards,

sc979

PR #7089, opened 2018-12-24
fix(media): Fix XSS vulnerability.

callapa:
Fix move and update features also.


sc979:
QA validation failed. XSS vulnerability is still present :'(

PR #7100, opened 2019-01-02
fix(secu): Token generation uses predictable generator

gquere:
Not really exploitable, but [uniqid()](http://php.net/uniqid) is a rather predictable generator, even with the added entropy.
When dealing with authentication it's better to use a cryptographically secure generator such as [openssl_random_pseudo_bytes](http://www.php.net/manual/en/function.openssl-random-pseudo-bytes.php) or [random_bytes()](http://php.net/manual/en/function.random-bytes.php).

### DIFF
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -93,7 +93,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST' &&
     }
 
     /* Insert Token in API webservice session table */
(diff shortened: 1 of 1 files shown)


victorvassilev:
thank you, it will be implemented with two pull requests to respective branches. closing this one with reference to: #7225 and #7224

PR #7124, opened 2019-01-14
Mon 3238 modules api details

vhr:
with changes from MON-3235


vhr:
the feature will be merged with epic branch

PR #7126, opened 2019-01-15
fix(widget) fix icon

loiclau:


www/include/home/customViews/views.php
```diff
@@ -146,12 +146,16 @@
...
+                jQuery('.widgetTitle').editable(
+                    './include/home/customViews/rename.php',
                     {
                         id: 'elementId',
                         name: 'newName',
                         event: 'dblclick'
-                    });
+                    }
+                );
+                var wrenchSpan = '<span class="ui-icon ui-icon-wrench"></span>';
```
kduret: I think those 2 variables should be initialized earlier.
Because if the use does not have permission, there will be an error

PR #7144, opened 2019-01-22
enh(core): add possibility to pass connection options to perl db lib

cgagnaire:
Add the possibility to pass connection options to perl DB handler connect function.
Useful to pass utf8 kind options.


garnier-quentin:
Can be found here: https://github.com/centreon/centreon/pull/7404

PR #7145, opened 2019-01-22
enh(core): enhance logger lib to handle utf8

cgagnaire:
Enhance handling of utf8 in perl logger lib.


garnier-quentin:
Merged with dbi connect options PR: https://github.com/centreon/centreon/pull/7404

PR #7173, opened 2019-02-04
enh(logs): Managing the removal of the log_id column from the centreon_storage.logs table

callapa:
This script will:
1. Stop broker
2. Rename the old log table (logs => logs_old)
3. Create a new log table without the log_id column (=> logs)
4. Restart Broker
5. Copy the data from the old log table to the new one

Pay attention, this script requires the root password of the database


tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,391 @@
...
+}
+touch($lockFileName);
+
+$currentStep = 1;
+try {
+    // We check if the database connection port is set, otherwise we request it.
+    if (defined('port')) {
+        $dbPort = port;
+    } else {
+        $port = askQuestion("Please enter the port of the Centreon database : ");
```
Pontissalien: In modern English-language printing, no space is placed before a colon
https://en.wikipedia.org/wiki/Colon_(punctuation)#Spacing

callapa: Fixed, I had forgotten this one.

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+
+/**
+ * Retrieve the started process number of Broker
+ *
+ * @return int Return the started process number
+ */
+function getBrokerProcess()
+{
+    exec(
+        'ps -o args -p $(pidof -o $$ -o $PPID -o %PPID -x cbd || echo 1000000) | grep -c /usr/sbin/cbd',
```
ganoze: You use _systemctl_ to start and stop the _cbd_ service. To be consistent you should use _systemctl status_ to detect if the service is running.

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+INTO TABLE logs 
+FIELDS TERMINATED BY ','
+LINES TERMINATED BY '\n'
+(ctime, host_id, host_name, instance_name, issue_id, msg_type, notification_cmd, 
+notification_contact, output, retry, service_description, service_id, status, type)
+QUERY;
+
+// We create a loc file to avoid to start this script one more time
+$fileInfos = pathinfo(__FILE__);
+$lockFileName = $fileInfos['filename'] . '.loc';
```
ganoze: IMHO _.lock_ would be more clear, but that is a detail.

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+
+    // And we rename the new table 'logs_new' to 'logs'
+    $logs("And we rename the new table 'logs_new' to 'logs'");
+    $db->query("ALTER TABLE logs_new RENAME TO logs");
+    $currentStep++;
+
+    $globalStep = $currentStep;
+
+    // We start Broker if it was previously started
+    if ($isBrokerAlreadyStarted) {
```
ganoze: I would have expected Centreon Broker to be restarted only at the end of the migration process. Otherwise some data could be inserted in the new table and getting this data back would be difficult if the migration has a failure and is aborted.

Note that this also includes table renaming. *logs* should remain named as such until the new table is 100% ready to receive new data.

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+
+        // We delete the csv temporary file
+        $logs("We delete the temporary csv file $partitionName");
+        unlink($pathPartition);
+        $currentStep++;
+
+        // If asked, we delete the old data from the old logs table
+        if ($shouldDeleteOldData) {
+            $logs("We delete the old data from the old logs table");
+            $db->query("DELETE FROM logs_old PARTITION ($partitionName)");
```
ganoze: Again, IMHO data should be dropped only after a successful migration.

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+    }
+
+    // We will process partitions that are not empty
+    $partitions = getNotEmptyPartitions($db);
+
+    if (! isset($shouldDeleteOldData)) {
+        // We ask to user if he want to keep old data otherwise we will delete the old logs table
+        $explanation = <<<TEXT
+Before starting, we inform you that we will create a new centreon_storage.logs table and rename the older.
+Next, we will copy data from the older table to new one\n\n
```
sc979:
```suggestion
Then, we will copy the data from the older table to the new one\n\n
```

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+TEXT;
+        printf($explanation);
+
+        $shouldDeleteOldData =
+            !askYesOrNoQuestion("Do you want to keep the old logs table ?");
```
sc979: To be consistent, shouldn't you add a space after the "!" ?

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+    }
+    // For consistency consideration, we stop broker before creating the new table
+    if ($isBrokerAlreadyStarted) {
+        $logs("For consistency consideration, we stop broker before creating the new table");
+        if (stopBroker() !== 0) {
+            throw new Exception("Error stopping Broker");
```
sc979:
```suggestion
            throw new Exception("Error stopping Broker. Please check Broker status");
```

sc979: Maybe the customer won't check if broker is still running

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+        }
+    } else {
+        $logs("Broker is not started, no need to stop it");
+    }
+    $currentStep++;
+
+    // We create the new logs table and rename older
```
sc979:
```suggestion
    // We create the new logs table and rename the older
```

tools/update_centreon_storage_logs.php
```diff
@@ -0,0 +1,387 @@
...
+    $logs("We create the new logs table and rename older");
```
sc979:
```suggestion
    $logs("We create the new logs table and rename the older");
```

tools/update_centreon_storage_logs.php
sc979: +1

tools/update_centreon_storage_logs.php
Pontissalien: And remove the space before the "?"
https://en.wikipedia.org/wiki/Question_mark#Stylistic_variants
> in the English language orthography no space is allowed in front of the question mark

ganoze:
Should be closed in favor of #7786 , right @callapa ?

PR #7177, opened 2019-02-05
Update centreon-backup-mysql.sh

pkriko:
Handle xfs file system just like the script sync-bigdb used for the failover.


garnier-quentin:
Already in PR https://github.com/centreon/centreon/pull/7242
So we close it

PR #6509, opened 2018-08-01
MON-3518-filter-correction-recurrent-downtime-form

sc979:
Resolve MON-3518 on 2.8.x
As I couldn't push a coding style commit, on jben94's branch. 
I opened a new branch to add the required commit.

From jben94 PR -> 
Link issue : #6501
Link to the first PR : #6504 closed / rebased on the master
Link to the second PR PR : #6506

PR #7201, opened 2019-02-19
enh(DT): fix search filter for recurrent downtimes

sc979:
Correct the number of rows to display when a filter is applied

### DIFF
--- a/www/class/centreonDowntime.class.php
+++ b/www/class/centreonDowntime.class.php
@@ -1,7 +1,7 @@
 <?php
 /**
- * Copyright 2005-2015 Centreon
- * Centreon is developped by : Julien Mathis and Romain Le Merlus under
+ * Copyright 2005-2019 Centreon
+ * Centreon is developed by : Julien Mathis and Romain Le Merlus under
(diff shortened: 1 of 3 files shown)


sc979:
this PR apply the 2.8.x fix on the 18.10.x :
https://github.com/centreon/centreon/pull/6509

www/class/centreonDowntime.class.php
```diff
@@ -116,13 +116,15 @@ public function getNbRows()
             return $this->nbRows;
         }
         /* Get the number of rows with a COUNT(*) */
-        $query = "SELECT COUNT(*) FROM downtime WHERE " . $this->search;
+        $query = "SELECT COUNT(*) FROM downtime" . ($this->search != '' ? ' WHERE :search' : '');
         try {
-            $res = $this->db->query($query);
+            $res = $this->db->prepare($query);
+            $res->bindValue(':search', $this->search, PDO::PARAM_STR);
```
kduret: you try to bind a value which does not exist if there is no search

PR #7202, opened 2019-02-19
fix(front): manage properly session expiration + avoid login inception

kduret:


adr-mo:
@kduret I have force pushed the branch for the validation in order to be **up to date** with the **18.10.x** branch :)

PR #7216, opened 2019-02-25
feat(build): set projectVersion 18.10.4 in sonar-project.properties.

kduret:


ganoze:
Currently the projectVersion is set within the mon-web-analysis script automatically. IMHO this PR should be rejected.

ganoze:
Similar to #7217

PR #7217, opened 2019-02-25
feat(build): set projectVersion 19.04.0 in sonar-project.properties.

kduret:


ganoze:
Currently the projectVersion is set within the mon-web-analysis script automatically. IMHO this PR should be rejected.

PR #7224, opened 2019-02-26
fix(api): update token generator                                     …

victorvassilev:
Updates the function to generate unpredictable token.
Resolves: MON-3311

<h2> Description </h2>

Updates the function to generate unpredictable token.

Fixes # 7224

<h2> Type of change </h2>

Please delete options that are not relevant.

- [x ] Patch fixing an issue (non-breaking change)

<h2> Target serie </h2>

- [ x] 2.8.x

Login to application and check the new token in database.

### DIFF
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -92,7 +92,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST' &&
     /**
      * generate cryptographically secure token
      */
-    require_once dirname(__FILE__) . '/class/hashGenerator.class.php';
+    require_once dirname(__FILE__) . '/class/cryptGenerator.class.php';
     $tokenHash = CryptGenerator::genBytes(32);
     $token = base64_encode($tokenHash);
     $res = $pearDB->prepare("INSERT INTO ws_token (contact_id, token, generate_date) VALUES (?, ?, NOW())");
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -88,7 +88,7 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST' &&
     }
 
     /* Insert Token in API webservice session table */
-    $token = base64_encode(uniqid('', true));
+    $token = base64_encode(random_bytes(32));
     $res = $pearDB->prepare("INSERT INTO ws_token (contact_id, token, generate_date) VALUES (?, ?, NOW())");
     $pearDB->execute($res, array($auth->userInfos['contact_id'], $token));
 
--- a/www/api/index.php
+++ b/www/api/index.php
@@ -88,7 +88,13 @@ if ($_SERVER['REQUEST_METHOD'] === 'POST' &&
     }
 
     /* Insert Token in API webservice session table */
-    $token = base64_encode(random_bytes(32));
+
+    /**
+     * generate cryptographically secure token
+     */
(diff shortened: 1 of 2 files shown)


www/api/index.php
```diff
@@ -88,7 +88,7 @@
     }
 
     /* Insert Token in API webservice session table */
-    $token = base64_encode(uniqid('', true));
+    $token = base64_encode(random_bytes(32));
```
sc979: random_bytes isn't supported in php 5.4

victorvassilev: you are completely right here! going to apply a fix for it, thanks

victorvassilev:
> Why bother rewriting a random number generator whereas the PHP documentation officially recommends using [this polyfill](https://github.com/paragonie/random_compat) ([source](http://php.net/manual/fr/function.random-bytes.php)) that is available as a Composer package ?

Because I don't want to introduce more dependencies for such a functionality that is already good enough. In theory we could also add openssl mod and use openssl_pseudo_bytes, but that is again overkill for the situation.

ganoze:
> Because I don't want to introduce more dependencies for such a functionality that is already good enough. In theory we could also add openssl mod and use openssl_pseudo_bytes, but that is again overkill for the situation.

I would object that the paragonie/random_compat package has no dependency besides PHP itself. Also whereas the final random number generator will be the same (/dev/urandom), the package performs more sanity check and is therefore far more robust IMHO (no offense intended).

Does anybody else have some opinion on this ?

sc979:
More security is always better... 
And as, if this pull request concerns only the api token generator, we can find uniqid() method called in other modules like : 
- openTicket
- broker's wizard
- installBaseConf script
- the upgrade script from 2.8.26 to 18.10.0
- the AppKeyGeneratorService's generateKey() method
and also in the QuickFormCustom's createSecurityToken() method
I think that a more secure solution is better ;)

victorvassilev:
The topic of this PR is to fix an improbable, but still possible, collision of unique hash generation in API token. Considering the state of the application as of 2.8.x version, and the implications to follow from introducing a new library, the provided solution so far resolves the problem and generates higher level of cryptographic unique token.

kduret:
This PR is only related to 2.8.x.
So, even the package performs more sanity check, it is too much impacting for this version to introduce this dependency. Cause currently, we don't use compose at all for 2.8.x

ganoze:
> This PR is only related to 2.8.x.
> So, even the package performs more sanity check, it is too much impacting for this version to introduce this dependency. Cause currently, we don't use compose at all for 2.8.x

Would a manual installation of the package work for you ?

adr-mo:
This patch is **way too impacting for the 2.8.x**

Closing the PR, the patch will only be applied on the **18.10.x and master**

ganoze:
@adr-mo Could you provide more details on why this is *way too impacting* ?

adr-mo:
@ganoze I've spoken with @kduret that told be that it was impacting. Maybe the _way too_ was too much.
However it seems that it does not worth the effort to integrate it in **2.8.x**

ganoze:
@kduret Could you explain what the technical impacts would be ?

kduret:
I asked @adr-mo to close it because we will waste too much time to validate something which is not really a security issue.
I consider this as an improvement, and IMO this kind of stuff should not be done on 2.8.x branch

PR #7250, opened 2019-03-06
feat(remote-server): allow usage of domain names

victorvassilev:
<h2> Description </h2>
Allows the user to define remote server by address not only by IP

Fixes # MON-3257

<h2> Type of change </h2>

Please delete options that are not relevant.

- [x ] Breaking change (patch or feature) that might cause side effects breaking part of the Software

<h2> Target serie </h2>
- [ x] 18.10.x

### DIFF
--- a/src/Centreon/Domain/Repository/InformationsRepository.php
+++ b/src/Centreon/Domain/Repository/InformationsRepository.php
@@ -71,9 +71,14 @@ class InformationsRepository extends ServiceEntityRepository
         $sql = "DELETE FROM `informations` WHERE `key` = 'authorizedMaster'";
         $stmt = $this->db->prepare($sql);
         $stmt->execute();
+
+        /*
+         * resolve the address down to IP
+         */
+        $ipAddress = gethostbyname($ip);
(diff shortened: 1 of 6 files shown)


src/CentreonRemote/Domain/Service/NotifyMasterService.php
```diff
@@ -36,13 +35,21 @@ class NotifyMasterService
     private $curl;
 
     /**
-     * @return Curl
+     * @return void
```
vhr: just remove  a **return** annotation

src/Centreon/Domain/Repository/InformationsRepository.php
```diff
@@ -71,9 +71,14 @@ public function authorizeMaster(string $ip): void
         $sql = "DELETE FROM `informations` WHERE `key` = 'authorizedMaster'";
         $stmt = $this->db->prepare($sql);
         $stmt->execute();
+
+        /*
+         * resolve the address down to IP
+         */
+        $ipAddress = gethostbyname($ip);
```
vhr: maybe this preparation should be done before IP is pass to the repository method

PR #7335, opened 2019-03-25
fix(ci): adding or increasing spin in service template acceptance tests

sc979:
<h1> Pull Request Template </h1>

<h2> Description </h2>

Increasing existing spin's time and adding a spin to the delete service template acceptance test

<h2> Type of change </h2>

<h2> Target serie </h2>

- [X] 19.04.x (master)

### DIFF
--- a/features/bootstrap/ServiceTemplateConfigurationContext.php
+++ b/features/bootstrap/ServiceTemplateConfigurationContext.php
@@ -330,7 +330,7 @@ class ServiceTemplateConfigurationContext extends CentreonContext
                     return count($this->tableau) == 0;
                 },
                 "Some properties are not being updated : ",
-                5
+                10
             );
         } catch (\Exception $e) {
             $this->tableau = array_unique($this->tableau);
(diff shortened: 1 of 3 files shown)


features/bootstrap/ServiceTemplateConfigurationContext.php
```diff
@@ -388,11 +388,21 @@ function ($context) {
      */
     public function iDeleteAServiceTemplate()
     {
-        $this->currentPage = new ServiceTemplateConfigurationListingPage($this);
-        $object = $this->currentPage->getEntry($this->initialProperties['description']);
-        $this->assertFind('css', 'input[type="checkbox"][name="select[' . $object['id'] . ']"]')->check();
-        $this->setConfirmBox(true);
-        $this->selectInList('select[name="o1"]', 'Delete');
+        try {
```
kduret: try is not useful, spin will throw the exception

features/bootstrap/ServiceTemplateConfigurationContext.php
```diff
@@ -388,11 +388,21 @@ function ($context) {
...
+            $this->spin(
```
ganoze: I do not understand why spin() is even needed. Deletion is supposed to be synchronous, therefore a spin() on the ServiceTemplateConfigurationListingPage should not be needed. Am I right ?

PR #7343, opened 2019-03-27
fix(graph): display thresholds in Split chart mode

### LINKED ISSUE #7235: plugin=network::juniper::mag::plugin thresholds on graphs
Hello, 

i was working with the users mode for plugin network::juniper::mag::plugin. 
Graphs are working but i can't see warning and critical thresholds lines. As you can see below, we can not see this on graph, the line is blue and keeps being blue even on critical status. 

image

I would like to have something like that: 
image

Tried to configure a template and a curve but it did not do the trick.

Can you help with this please ?

Regards

### LINKED ISSUE #7342: Graphs in "Split chart" mode do not show thresholds
Hi,

When a graph contains several curves, thresholds are not displayed (rather understandable, even in some cases they could be displayed).

However, when in "Split chart" mode, so with only one curve per graph, thresholds are not displayed either.

I think that in that case, they should be displayed.

Thank you 👍

Edit : issue still present in 19.04.0.

CPbN:
Hi,

<h2> Description </h2>

This PR activates thresholds display on graphs when in Split chart mode, so when there is only one curve per graph.

Fixes #7342
Should also close #7235.

<h2> Type of change </h2>

<h2> Target serie </h2>

- [x] 18.10.x
- [x] 19.04.x (master)

Display a graph with several curves, and with at least with thresholds.
With all curves on the same graph, thresholds are not displayed.
Go in Split chart mode, and verify here that thresholds are correctly shown.

Thank you 👍

### DIFF
--- a/www/class/centreonGraph.class.php
+++ b/www/class/centreonGraph.class.php
@@ -331,6 +331,9 @@ class CentreonGraph
         } elseif ($metrics != "") {
             $this->metricsEnabled = array($metrics);
         }
+        if (is_array($this->metricsEnabled) && (count($this->metricsEnabled) == 1)) {
+            $this->onecurve = true;
+        }
     }
 
     /**


adr-mo:
Hi @CPbN,

Thanks for the contribution :100: 

We'll try to make your work happen in Centreon as soon as possible :+1: 

Regards,

adr-mo:
Creating a new branch from yours @CPbN to do the validation :)

CPbN:
Perfect @adr-mo, feel free, many thx :)

PR #7347, opened 2019-03-28
fix(install) Get the ip address of an existing connection to set the permission correctly

lgcosta:
<h1> Pull Request Template </h1>

<h2> Description </h2>

When an installation of the Centreon is done using the database server (MySQL) on an external server and it is on another network being accessed only via NAT, an ip address is mistakenly assigned in the user permission `centreon` of MySQL.

Currently it uses a server variable to get the client ip (`$_SERVER['SERVER_ADDR']`). however, when the server is in another network location, the ip that needs to be allocated in the user's permission is the network gateway (which does the NAT).

As for getting to this `centreon` user creation function, it is necessary to have the connection using valid `root`, I am taking into account that there is a valid connection and functioning as `root`. ~~So, I have a attribute available in the PDO connection object and with that, I use an attribute available `PDO::ATTR_CONNECTION_STATUS` in the object to get the valid ip address of the client for the connection.~~

Using mysql directly is more effective, the PDO was returning me the ip of the server.

```
mysql> select host from information_schema.processlist WHERE ID=connection_id();
+------------------+
| host             |
+------------------+
| 172.17.0.1:45842 |
+------------------+
```
This way, I use the Mysql information table and I can capture the ip of the current session.

<h2> Type of change </h2>

<h2> Target serie </h2>

- [ ] 2.8.x
- [x] 18.10.x
- [x] 19.04.x (master)

I came up with this problem using environments where I used MariaDB on docker instances in networks other than the Centreon installation. For example, I have the MariaDB docker on the 172.17.0.0/24 network and the Centreon installation on a Vagrant on the 10.1.0.0/24 network.

<h2> Checklist </h2>

<h5> Community contributors & Centreon team </h5>

- [ ] I have **rebased** my development branch on the base branch (master, maintenance).

<h5> Centreon team only </h5>

- [ ] I have made sure that the **unit tests** related to the story are successful.
- [ ] I have made sure that **unit tests covers 80%** of the code written for the story.
- [ ] I have made sure that **acceptance tests** related to the story are successful (**local and CI**)

### DIFF
--- a/www/install/steps/process/createDbUser.php
+++ b/www/install/steps/process/createDbUser.php
@@ -64,7 +64,7 @@ $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
     $getIpQuery = $link->prepare(
-        'SELECT host FROM information_schema.processlist WHERE ID=connection_id()'
+        'SELECT host FROM information_schema.processlist WHERE ID = connection_id()'
     );
     $getIpQuery->execute();
     // The result example (172.17.0.1:38216), use the explode function to remove port
--- a/www/install/steps/process/createDbUser.php
+++ b/www/install/steps/process/createDbUser.php
@@ -64,7 +64,7 @@ $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
     $getIpQuery = $link->prepare(
-        'select host from information_schema.processlist WHERE ID=connection_id()'
+        'SELECT host FROM information_schema.processlist WHERE ID=connection_id()'
     );
     $getIpQuery->execute();
     // The result example (172.17.0.1:38216), use the explode function to remove port
--- a/www/install/steps/process/createDbUser.php
+++ b/www/install/steps/process/createDbUser.php
@@ -63,7 +63,7 @@ $dbPass = $parameters['db_password'];
 $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
-    $host = $_SERVER['SERVER_ADDR'];
+    $host = explode(" ", $link->getAttribute(PDO::ATTR_CONNECTION_STATUS))[0];
 }
 $query = "GRANT ALL PRIVILEGES ON `%s`.* TO `" . $dbUser . "`@`" . $host .
     "` IDENTIFIED BY '" . $dbPass . "' WITH GRANT OPTION";
--- a/www/install/steps/process/createDbUser.php
+++ b/www/install/steps/process/createDbUser.php
@@ -63,7 +63,12 @@ $dbPass = $parameters['db_password'];
 $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
-    $host = explode(" ", $link->getAttribute(PDO::ATTR_CONNECTION_STATUS))[0];
+    $getIpQuery = $link->prepare(
+        'select host from information_schema.processlist WHERE ID=connection_id()'
+    );
+    $getIpQuery->execute();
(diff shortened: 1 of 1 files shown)


www/install/steps/process/createDbUser.php
```diff
@@ -63,7 +63,7 @@
 $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
-    $host = $_SERVER['SERVER_ADDR'];
+    $host = explode(" ", $link->getAttribute(PDO::ATTR_CONNECTION_STATUS))[0];
```
ganoze: The value returned by *PDO::ATTR_CONNECTION_STATUS* seems to be undefined to say the least.

What do you think about reworking our hostname retrieval with a **SELECT USER()** query ? This function is defined for both [MySQL](https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_user) and [MariaDB](https://mariadb.com/kb/en/library/user/).

lgcosta: Yes @ganoze I was verifying exactly that, it was much more effective I directly query mysql. I ended up doing a little different from what he suggested, but in the same way.
Thanks

www/install/steps/process/createDbUser.php
```diff
@@ -63,7 +63,12 @@
 $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
-    $host = explode(" ", $link->getAttribute(PDO::ATTR_CONNECTION_STATUS))[0];
+    $getIpQuery = $link->prepare(
+        'select host from information_schema.processlist WHERE ID=connection_id()'
```
ganoze: Please fix the syntax of the SQL query for consistency (keywords in upper case, identifiers in lower case) and that will be all good for me.

lgcosta: done. Thanx

www/install/steps/process/createDbUser.php
```diff
@@ -63,7 +63,12 @@
 $host = "localhost";
 // if database server is not on localhost...
 if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhost") {
-    $host = $_SERVER['SERVER_ADDR'];
+    $getIpQuery = $link->prepare(
+        'SELECT host FROM information_schema.processlist WHERE ID=connection_id()'
```
sc979:
```suggestion
        'SELECT host FROM information_schema.processlist WHERE ID = connection_id()'
```

sc979: Hi Igcosta, 
Please, can you add spaces before and after the equal. As, now, sonar broke every job on jenkins when new coding style issue are added.

lgcosta:
@kduret, @ganoze, @sc979 can I merge this PR?

kduret:
is it possible that "SELECT host FROM information_schema.processlist WHERE ID = connection_id()" does not return anything ?
does it respect our MySQL/MariaDB compatibilities ?

lgcosta:
> is it possible that "SELECT host FROM information_schema.processlist WHERE ID = connection_id()" does not return anything ?
> does it respect our MySQL/MariaDB compatibilities ?

- The `information_schema` is a core table from Mysql, common in Mysql and Mariadb [1] and [2]
- The function `connection_id()` is a core function, common in Mysql and Mariadb [3] and [4]

As it takes advantage of an active match from the `root` connection, the field will return a value in relation to an active `connection_id()`. This is only different if I connect a client directly to the Mysql socket file, the field will return empty (I did the test). However, we do not have the case of direct connection on the socket, we always do over inet (tcp).

[1] https://dev.mysql.com/doc/refman/8.0/en/processlist-table.html
[2] https://mariadb.com/kb/en/library/information-schema-processlist-table/
[3] https://dev.mysql.com/doc/refman/8.0/en/information-functions.html#function_connection-id
[4] https://mariadb.com/kb/en/library/connection_id/

kduret:
Ok thanks for reply. Then you can merge it ;)

PR #7371, opened 2019-04-01
fix(UI): correctly toggle edit when widgets load

CPbN:
Hi,

<h2> Description </h2>

When widgets page load, edit mode is disabled, but every widget edit mode remains enabled, as shown in the screenshot below. This PR then fixes this.

55336917-01023a80-549e-11e9-940c-2866a4f4099c

<h2> Type of change </h2>

<h2> Target serie </h2>

- [x] 19.10.x
- [x] 20.04.x (master)

Display widgets pages and be sure edit buttons are hidden for every widget.

Thank you 👍

Edit : issue still present in 19.10.1.

### DIFF
--- a/www/include/home/customViews/index.ihtml
+++ b/www/include/home/customViews/index.ihtml
@@ -869,6 +869,9 @@ jQuery(function () {
                 window.location.reload();
             }
           },
+          load: function( event, ui ) {
+              toggleEdit(defaultShow);
+          },
           activate: function( event, ui ) {
             tabLoops++;
             columnClass = null;
--- a/www/include/home/customViews/index.ihtml
+++ b/www/include/home/customViews/index.ihtml
@@ -579,7 +579,7 @@ function toggleEdit(show) {
     jQuery('#actionBar').hide();
     jQuery('.portlet-header').each(function(index) {
       var widgetTitle = jQuery(this).find('.widgetTitle').text().trim();
-      if (widgetTitle == 'Click to edit' || widgetTitle == '') {
+      if (widgetTitle == 'Click to edit') {
         jQuery(this).hide();
       } else {
         jQuery(this).find('.ui-icon-wrench').hide();
--- a/www/include/home/customViews/index.ihtml
+++ b/www/include/home/customViews/index.ihtml
@@ -582,7 +582,8 @@ function toggleEdit(show) {
       if (widgetTitle == 'Click to edit' || widgetTitle == '') {
         jQuery(this).hide();
       } else {
-        jQuery(this).find('.ui-icon').hide();
+        jQuery(this).find('.ui-icon-wrench').hide();
+        jQuery(this).find('.ui-icon-trash').hide();
       }
     });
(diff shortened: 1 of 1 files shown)


jdelpierre:
Hi :) 

Do you speak the "wrench" icon (ui-icon ui-icon ui-icon-wrench) next to the "trash" icon, on the widget headers when the page is in "reading"?
Should we hide it when we are not in "editing"?

CPbN:
I speak about the all 4 icons at the top right of every widget.
They correctly hide when we leave edition mode.
But when we load the widget page (when we enter the Home / Csutom Views page), they show-up by default, even if edition mode is off by default.
This PR then corrects this, correctly toggling edition mode when tabs are fully loaded.

CPbN:
Hi @jdelpierre :)
I speak about the all 4 icons at the top right of every widget.
They correctly hide when we leave edition mode.
But when we load the widget page (when we enter the Home / Custom Views page), they show-up by default, even if edition mode is off by default.
This PR then corrects this, toggling edition mode when tabs are fully loaded, which then correctly sets icons' state.

CPbN:
Many thx @jdelpierre & @sc979 :)

lpinsivy:
@CPbN

Good be good to don't hide "reload" and "minimize" icons in non edition mode

Regards

CPbN:
Done @lpinsivy :)
Many thanks for your feedback :+1:

CPbN:
@lpinsivy you should be able to approve this as I applied your modification request :)
Thank you 👍

CPbN:
Still facing this in 19.10.0-beta.2, let's then merge this tiny PR please ?
Thx !

CPbN:
Wigdets without title hide the buttons, certainly to save some vertical space.
I did not change this behavior (I like it actually :wink:)
Put some dummy title to show the buttons, if you need them ?

loiclau:
We will discuss of this asap

www/include/home/customViews/index.ihtml
```diff
@@ -582,7 +582,8 @@ function toggleEdit(show) {
       if (widgetTitle == 'Click to edit' || widgetTitle == '') {
         jQuery(this).hide();
```
CPbN: @loiclau we could also remove the `widgetTitle == ''` condition, so that, to hide the title bar, we would simply have to let the "Click to edit" default title.
If this suits your needs, I can push this modification.
Thx 👍

loiclau:
Hi @CPbN , 
finally we want to keep the "reload" and "minimize" icon when the widget have no title in non edition mode

CPbN:
Hi @loiclau, thank you for your feedback, I then just pushed a commit accordingly.

loiclau:
I think you can delete the if condition to get always the reload option.

CPbN:
Actually this condition gives a small trick to be able to hide the title bar.
One may like to hide it to save some vertical space.
Think about for example a dashboard displayed on a wall-mounted screen, no-one has access to this reload button, so it may be a good thing to hide it.
Thx !

PR #7390, opened 2019-04-05
update placeholder path replacer to use sed

victorvassilev:
*Description*
Updates php replacer to replace generated FE static files placeholders better
*Type of change*

• [ x] Breaking change (patch or feature) that might cause side effects breaking part of the Software

*Target serie*

• [ x] 19.10

*How this pull request can be tested ?*

make npm run build in main folder and then access index.php, if frontend shows properly - the replacement was successful

*Checklist*
*Community contributors & Centreon team*

☑︎ I followed the *coding style guidelines* provided by Centreon
☑︎ I have commented my code, especially new *classes*, *functions* or any *legacy code* modified. (_*docblock*_)
☑︎ I have commented my code, especially *hard-to-understand areas* of the PR.
☑︎ I have made corresponding changes to the *documentation*.
• [x] I have *rebased* my development branch on the base branch (master, maintenance).

### DIFF
--- a/www/index.php
+++ b/www/index.php
@@ -101,10 +101,8 @@ if (!$staticExists) {
     $allJsFiles = glob('static/js/*');
     $indexFile = glob('index.html');
     $allFiles = array_merge($allCssFiles, $allJsFiles, $indexFile);
-    foreach ($allFiles as $file) {
-        $fc = file_get_contents($file);
-        $newCont = str_replace('_CENTREON_PATH_PLACEHOLDER_', $newPath, $fc);
-        file_put_contents($file, $newCont);
+    foreach ($allFiles as $file){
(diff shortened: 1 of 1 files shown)


victorvassilev:
That is precisely why I replaced it, the str_replace didn't work well on one of the npm generated files, I haven't gone in detail why, but this solution works and introduces a quick fix before new release. Right now it is quite possible a build of the new frontend breaks the system. I am up for opening a story and pursuing this further as php implementation, but unfortunately, we are focused on another topic right now.

sc979:
I wont agree with this changes as I agree with @ganoze.
Indeed it's late to rework the frontEnd part, but I've told you about npm dependencies problems.
Why wait now, to find a "quick fix" when you could find a better way before ?
Is a "quick fix" better than a "maintainable dev" ?
Are you sure there will be no border effect when we'll need to make this "quick fix" a maintainable one ?
I'll follow @kduret decision...

PR #7502, opened 2019-04-29
enh(sec): log failed login attempts (ban post processing)

CPbN:
Hi,

<h2> Description </h2>

This PR simply logs failed login attempts.
Goal is then for the system administrator to be able to use tools such as fail2ban to ban IPs generating too many failed login attempts.
As fail2ban generally already analyzes Apache error log file, these error messages are directly sent to it (thanks to the second option of `error_log` set to `4`).

<h2> Type of change </h2>

<h2> Target serie </h2>

- [x] 19.04.x
- [x] 19.10.x (master)

`tail -f /var/log/httpd24/error_log`
Make failed login attempts (easy) !

Thank you :+1: 

Edit : still missing from 19.10.0-beta.2.

### DIFF
--- a/www/class/centreonAuth.class.php
+++ b/www/class/centreonAuth.class.php
@@ -348,7 +348,10 @@ class CentreonAuth
                 $this->CentreonLog->insertLog(1, "[".$this->source."] No contact found with this login : '$username'");
             }
             $this->error = _('Your credentials are incorrect.');
-            error_log("user " . $username . " not found", 4);
+            //Strangely fresh installs reach this code with empty username when hitting the portal
+            if (strlen($username) > 0) {
+                error_log("user " . $username . " not found", 4);
+            }
         }
     }
 
--- a/www/class/centreonAuth.class.php
+++ b/www/class/centreonAuth.class.php
@@ -305,6 +305,7 @@ class CentreonAuth
                 } else {
                     $this->CentreonLog->insertLog(1, "Contact '" . $username . "' doesn't match with password");
                     $this->error = _('Your credentials are incorrect.');
+                    error_log("user " . $username . " password mismatch", 4);
                 }
             } else {
                 $this->CentreonLog->insertLog(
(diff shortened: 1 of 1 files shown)


CPbN:
Many thanks @sc979 :+1:

kduret:
It seems that fail2ban is able to read other files.
So it should be more relevant to read this file : /var/log/centreon/login.log
Instead of adding another line in apache logs

CPbN:
Goal is also to avoid increasing number of files to read from Fail2Ban, to avoid load impact.
As Fail2Ban already reads the Apache error log file (of course if admin enables the Apache jail), then the error messages from this PR will be caught without having to process another file.
I could tune message a little bit to tell that they come from Centreon :)

CPbN:
Would be nice to have this in 19.10 ?
Thank you !

kduret:
@lpinsivy I guess you can confirm, but we will not add applicative logs in apache logs.
sorry for that @CPbN

CPbN:
@kduret that's a pity :confused:
I think Apache error log (the one targeted by this PR) is the good place to log such error messages :

- if you install Fail2Ban, you will for sure enable the Apache jail for common Apache rules, you will then already check Apache error log file : we then minimize load impact and simplify config ;
- Apache already sends _authentication failed_ error messages into its error log file, for example when HTTP basic authentication is used : makes then sense for HTTP applications (such as Centreon) to send their _authentication failed_ error messages to this error log file.

In addition, and for security reasons, this PR is a good thing to have, to protect against unauthorised access attempts. Think about Centreon servers reachable through Internet / public networks for example...

May I then ask you to reconsider this PR, as per the previous various arguments ?

Thank you again for your support 👍

CPbN:
@kduret @lpinsivy did you manage to double-check this ?
Thank you very much again 👍

kduret:
HTTP basic authentication is directly related to HTTP, so it makes sense to put it in apache logs.
But Centreon authentication is related to Centreon web,  and should not be part of apache logs (even if configuration of fail2ban is more complex)

So, sorry for that, but we will definetely not merge this PR.

CPbN:
Thank you @kduret for your additional feedback.
What would then be the correct destination for such messages ?
An existing log file ?
A new one (`/var/log/centreon/login.log` as you seemed to propose above) ?

I can submit a new PR with your desired target.
Thank you !

CPbN:
Thank you @kduret for your additional feedback.
Then see #7943 for another solution which should suit you :+1:

PR #7583, opened 2019-06-04
fix(clapi): remove carriage returns from string values

victorvassilev:
## Description

Remove carriage returns from strings in response

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [x] 19.04.x
- [ ] 19.10.x (master)

Check #BAM-292


www/api/class/centreon_clapi.class.php
```diff
@@ -230,4 +235,12 @@ public function authorize($action, $user, $isInternal = false)
 
         return $user->hasAccessRestApiConfiguration();
     }
+
+    /**
+     * Removes carriage returns from $item if string
+     * @param $item variable to check
```
callapa:
```suggestion
     * @param mixed &$item variable to check
```

www/api/class/centreon_clapi.class.php
```diff
@@ -230,4 +235,12 @@ public function authorize($action, $user, $isInternal = false)
...
+     */
+    private function clearCarriageReturns(&$item) {
```
callapa: The name of the method is confusing because it does not only remove carriage returns (\t)

PR #7598, opened 2019-06-10
Add ESLint package and generate configuration.

bdauria:
## Description

This PR sets up the ESLint configuration, extending the airbnb style, and using the prettier integration.
The next step is to re-format the code-base using the new rules.


.eslintrc.js
```diff
@@ -0,0 +1,22 @@
+module.exports = {
```
bdauria: We might want to start sharing the common config accross centreon & centreon-react-components. I'm not sure what's the best way to go about it. Put it in centreon-react-components and consume it in centreon?

bdauria: one approach for ESLint would be to create a config plugin: https://eslint.org/docs/developer-guide/shareable-configs

PR #7642, opened 2019-06-20
Enable Broker stats on Remote Server

### LINKED ISSUE #7635: [Remote Server] No Broker Statistics (debug / investigation)
Hi,

When you configure a Remote Server, Broker Statistics ( in _Administration / Platform Status / Broker Statistics_) are not available.

I investigated, and found that on the Remote Server, in `cfg_centreonbroker` table, `config_filename` field do not point to the correct filenames.
If I manually update them to the proper filenames, Broker Statistics correctly show.
Unfortunately these fields are updated on every poller configuration export.

Let's go further.

`export()` function from `PollerExporter.php` triggers to export the broker configuration.
And the configuration get messed-up by the following call :
`$cfgCentreonBroker = $overwritePollerService->getCfgCentreonBroker...`
`getCfgCentreonBroker()` calls `findPollerAndSetResourceData()` which deliberately removes the Remote Server configuration and replaces it by a default configuration template.

I don't understand why... But this does not seem to be correct...
Of course as `findPollerAndSetResourceData` is called for other configuration items (variables, engine...) impact may be bigger.

Coud you please investigate this further ?

Thank you 👍

Edit : issue still present in 19.10.0-beta.2.

CPbN:
Hi,

## Description

On a Remote Server, broker stats are not available.
This PR fixes this.

Fixes #7635

## Type of change

## Target serie

- [x] 19.04.x
- [x] 19.10.x (master)

Go to a Remote Server and be sure broker stats are available.

Thx 👍

Edit : issue still present in 19.04.2.

### DIFF
--- a/src/CentreonRemote/Domain/Resources/RemoteConfig/CfgCentreonBroker.php
+++ b/src/CentreonRemote/Domain/Resources/RemoteConfig/CfgCentreonBroker.php
@@ -23,7 +23,21 @@ class CfgCentreonBroker
             'broker' => [
                 'config_name'            => "{$configName}-broker",
                 'config_filename'        => "central-broker.xml",
-                'config_write_timestamp' => '0',
+                'config_write_timestamp' => '1',
+                'config_write_thread_id' => '0',
+                'config_activate'        => '1',
+                'ns_nagios_server'       => $serverID,
+                'event_queue_max_size'   => '100000',
+                'command_file'           => '',
+                'cache_directory'        => '/var/lib/centreon-broker',
+                'stats_activate'         => '1',
+                'correlation_activate'   => '0',
+                'daemon'                 => '1',
+            ],
+            'rrd' => [
+                'config_name'            => "{$configName}-rrd",
+                'config_filename'        => "central-rrd.xml",
+                'config_write_timestamp' => '1',
                 'config_write_thread_id' => '0',
                 'config_activate'        => '1',
                 'ns_nagios_server'       => $serverID,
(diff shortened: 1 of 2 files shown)


CPbN:
Mmmh I forgot to mention that I also had to rename the Remote Server broker configuration files, in the broker configuration, to the default ones :
- central-broker.xml
- central-rrd.xml
- central-module.xml

This to match the `cfg_centreonbroker` table on the Remote Server...
So this may only part of the solution... Or simply an idea to solve the issue...

CPbN:
My last commit should do the trick :+1:

CPbN:
While here, I also aligned the Remote broker configuration to the [Default one](https://github.com/centreon/centreon/blob/master/src/CentreonRemote/Domain/Resources/DefaultConfig/CfgCentreonBroker.php).

lpinsivy:
Hi @CPbN I'm not sure that your solution is correct because you hard coded the name of the file.

For me the best way it to export the correct name for the broker XML configuration file name of a Remote Server.

CPbN:
> Hi @CPbN I'm not sure that your solution is correct because you hard coded the name of the file.

Yes, this to mimic the default configuration.
And as we have only one broker on a Remote Server, I'm not sure this is dangerous.

> For me the best way it to export the correct name for the broker XML configuration file name of a Remote Server.

Yes I also think this is the other possible solution :)
I did not went through this part of the code to see whether or not this is difficult to achieve.

CPbN:
See #7644 for another solution which should be better :)

PR #7661, opened 2019-07-01
fix(security): fix contact permissions if user is not administrator

miteto:
# fix contact permissions if user is not administrator

REFS: MON-3200

## Type of change

## Target serie

- [ ] 2.8.x
- [X] 18.10.x
- [X] 19.04.x
- [X] 19.10.x (master)

### DIFF
--- a/www/include/configuration/configObject/contact/formContact.php
+++ b/www/include/configuration/configObject/contact/formContact.php
@@ -1,7 +1,7 @@
 <?php
 /*
  * Copyright 2005-2019 Centreon
- * Centreon is developped by : Julien Mathis and Romain Le Merlus under
+ * Centreon is developed by : Julien Mathis and Romain Le Merlus under
  * GPL Licence 2.0.
  *
  * This program is free software; you can redistribute it and/or modify it under
@@ -73,15 +73,15 @@ $initialValues = array();
 /*
  * Check if this server is a Remote Server to hide some part of form
  */
-$DBRESULT = $pearDB->query("SELECT i.value FROM informations i WHERE i.key = 'isRemote'");
-$result = $DBRESULT->fetchRow();
+$dbResult = $pearDB->query("SELECT i.value FROM informations i WHERE i.key = 'isRemote'");
+$result = $dbResult->fetchRow();
 if ($result === false) {
     $isRemote = false;
 } else {
     $isRemote = array_map("myDecode", $result);
     $isRemote = ($isRemote['value'] === 'yes') ? true : false;
 }
-$DBRESULT->closeCursor();
+$dbResult->closeCursor();
 
 $cct = array();
 if (($o == "c" || $o == "w") && $contact_id) {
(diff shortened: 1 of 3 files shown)


www/include/configuration/configObject/contact/formContact.php
```diff
@@ -92,10 +92,10 @@
     $cct["contact_svNotifCmds"] = array();
     $cct["contact_cgNotif"] = array();
 
-    $DBRESULT = $pearDB->query("SELECT * FROM contact WHERE contact_id = '" . intval($contact_id) . "' LIMIT 1");
-    $cct = array_map("myDecode", $DBRESULT->fetchRow());
+    $dbResult = $pearDB->query("SELECT * FROM contact WHERE contact_id = '" . intval($contact_id) . "' LIMIT 1");
```
vhr: Please use **bindValue** to set **contact_id** value
```suggestion
    $dbResult = $pearDB->prepare("SELECT * FROM contact WHERE contact_id = :contact_id LIMIT 1");
    $dbResult->bindValue(':contact_id', $contact_id, \PDO::PARAM_INT);
    $dbResult->execute();
```

www/include/configuration/configObject/contact/formContact.php
```diff
@@ -73,15 +73,15 @@
 /*
  * Check if this server is a Remote Server to hide some part of form
  */
-$DBRESULT = $pearDB->query("SELECT i.value FROM informations i WHERE i.key = 'isRemote'");
-$result = $DBRESULT->fetchRow();
+$dbResult = $pearDB->query("SELECT i.value FROM informations i WHERE i.key = 'isRemote'");
+$result = $dbResult->fetchRow();
```
sc979: please replace fetchRow with fetch
```suggestion
$result = $dbResult->fetch();
```

www/include/configuration/configObject/contact/formContact.php
```diff
@@ -92,10 +92,10 @@
     $cct["contact_svNotifCmds"] = array();
     $cct["contact_cgNotif"] = array();
 
-    $DBRESULT = $pearDB->query("SELECT * FROM contact WHERE contact_id = '" . intval($contact_id) . "' LIMIT 1");
-    $cct = array_map("myDecode", $DBRESULT->fetchRow());
+    $dbResult = $pearDB->query("SELECT * FROM contact WHERE contact_id = '" . intval($contact_id) . "' LIMIT 1");
+    $cct = array_map("myDecode", $dbResult->fetchRow());
```
sc979: fetchRow
```suggestion
    $cct = array_map("myDecode", $dbResult->fetch());
```

sc979:
Please remove the reference to an issue not linked to your fix and closed some years ago ;)
image

www/include/configuration/configObject/contact/formContact.php
```diff
@@ -784,6 +784,11 @@ function myReplace()
 
 if ($form->validate() && $from_list_menu == false) {
     $cctObj = $form->getElement('contact_id');
+    if (!$centreon->user->admin && $contactId) {
```
sc979: $contactId is never set... please can you rename all $centreon_id with $centreonId

miteto:
Why there is such a field (issue #) then ?

www/include/configuration/configObject/contact/formContact.php
miteto: contactId is set in contact.php, otherwise whole edit functionality would have been broken.

www/include/configuration/configObject/contact/formContact.php
```diff
@@ -784,9 +786,14 @@ function myReplace()
 
 if ($form->validate() && $from_list_menu == false) {
     $cctObj = $form->getElement('contact_id');
+    if (!$centreon->user->admin && $contact_id) {
```
sc979: you should have replaced all $contact_id with $contactId to respect the camelCase rule :'(

www/include/configuration/configObject/contact/formContact.php
```diff
@@ -784,9 +786,14 @@ function myReplace()
...
+        $form->getElement('contact_admin')->setValue(0);
```
sc979: I'm not sure that you should replace the value after the form submit, as you may overwrite values when you shouldn't -> eg : contactA has the right to reach API, but contactB interraction removes the right when the form is submitted.
You may need to hidden these fields when the form is displayed instead of overwritting the value set...

PR #7688, opened 2019-07-09
fix(TA): sort order when exporting using Clapi

sc979:
# Pull Request Template

## Description

CI / Jenkins jobs :
Trying to find a way to avoid random import order after an export in the Clapi.feature acceptance test


sc979:
Fixed by @loiclau ;)
Thanks

PR #7720, opened 2019-07-22
enh(traps): Add search options for statuses and vendors (manufacturers)

miteto:
## Description

Add search options for statuses and vendors (manufacturers)
(edit @sc979 in  Configuration > SNMP Traps > SNMP Traps listing page)

## Type of change

- [X] New functionality (non-breaking change)

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [X] 19.04.x
- [X] 19.10.x (master)

### DIFF
--- a/www/include/configuration/configObject/traps/listTraps.ihtml
+++ b/www/include/configuration/configObject/traps/listTraps.ihtml
@@ -13,7 +13,7 @@
         <tr>
             <td><input type="text" name="searchT" value="{$searchT}"></td>
             <td>
-                <select name="status" id="status">
+                <select class="select2" name="status" id="status">
                     <option value="">{t}All{/t}</option>
                     {foreach from=$statuses key=key item=status}
                         <option {if $searchStatus !="" && $key == $searchStatus}selected="selected"{/if}
@@ -22,7 +22,7 @@
                 </select>
             </td>
             <td>
-                <select name="vendor" id="vendor">
+                <select class="select2" name="vendor" id="vendor">
                     <option value="">{t}All{/t}</option>
                     {foreach from=$vendors item=vendor}
                         <option {if $vendor.id == $searchVendor}selected="selected"{/if}
@@ -96,3 +96,13 @@
     <input type='hidden' id='limit' name='limit' value='{$limit}'>
     {$form.hidden}
 </form>
+<script>
+    {literal}
+        $('.select2').centreonSelect2({
+            multiple: false,
+            select2: {
+                allowClear: false
+            }
+        });
+    {/literal}
+</script>
--- a/www/include/configuration/configObject/traps/listTraps.ihtml
+++ b/www/include/configuration/configObject/traps/listTraps.ihtml
@@ -16,7 +16,7 @@
                 <select class="select2" name="status" id="status">
                     <option value="">{t}All{/t}</option>
                     {foreach from=$statuses key=key item=status}
-                        <option {if $searchStatus !="" && $key == $searchStatus}selected="selected"{/if}
+                        <option {if $searchStatus != "" && $key == $searchStatus}selected="selected"{/if}
                             value="{$key}">{$status}</option>
                     {/foreach}
                 </select>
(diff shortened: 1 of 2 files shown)


www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -217,6 +243,8 @@ function setO(_i) {
 
 $tpl->assign('limit', $limit);
 $tpl->assign('searchT', $search);
+$tpl->assign('searchVendor', $searchVendor);
+$tpl->assign('searchStatus', (string)$searchStatus);
```
sc979: line 58, you're sanitizing the variable and searching for an integer. Why casting a string here ?

www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -53,25 +53,42 @@
...
+    $_POST['status'] ?? $_GET['status'] ?? null,
+    FILTER_SANITIZE_NUMBER_INT
+);
+
+$searchVendor = filter_var(
+    $_POST['vendor'] ?? $_GET['vendor'] ?? null,
+    FILTER_SANITIZE_NUMBER_INT
+);
+
 if (isset($_POST['searchT']) || isset($_GET['searchT'])) {
```
sc979: here you can use $search from line 51  to avoid to check the POST and GET again

www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -86,6 +103,14 @@
 $lvl_access = ($centreon->user->access->page($p) == 1) ? 'w' : 'r';
 $tpl->assign('mode_access', $lvl_access);
 
+// Get vendors (manufacturers) for search dropdown
+$vendorResult = $pearDB->query("SELECT * FROM traps_vendor ORDER BY name, alias");
```
sc979: if you need only the 'id' and 'name' column, don't pull every data from this table for better performances

www/include/configuration/configObject/traps/listTraps.ihtml
```diff
@@ -1,78 +1,98 @@
...
+            <td><h4>{t}Status{/t}</h4></td>
+            <td><h4>{t}Snmp trap manufacturer{/t}</h4></td>
+        </tr>
+        <tr>
+            <td><input type="text" name="searchT" value="{$searchT}"></td>
+            <td>
+                <select name="status" id="status">
+                    <option value="">{t}All{/t}</option>
+                    {foreach from=$statuses key=key item=status}
+                        <option {if $searchStatus !="" && $key == $searchStatus}selected="selected"{/if}
```
sc979:
```suggestion
                        <option {if $searchStatus != "" && $key == $searchStatus}selected="selected"{/if}
```

sc979:
Please could you correct the lowercase SQL query line 139 too (it makes my eyes bleed :))) )

www/include/configuration/configObject/traps/listTraps.php
miteto: it needs to be string in smarty, because there is status with value 0 and if the var is not string - it is always selected.

www/include/configuration/configObject/traps/listTraps.php
```diff
@@ -53,15 +53,20 @@
...
+    $searchVendor = filter_var(
+        $_POST['vendor'] ?? $_GET['vendor'] ?? null,
+        FILTER_SANITIZE_NUMBER_INT
+    );
+} else {
+    $searchStatus = null;
+    $searchVendor = null;
+}
 
 if (isset($_POST['searchT']) || isset($_GET['searchT'])) {
```
sc979: same comment. If the customer wich to select only a vendor, shouldn't we save its presets ?

sc979: maybe you could concatenate this block with the next (line 84) by replacing the line 71 with the line 84 condition

loiclau:
#7758

PR #7731, opened 2019-07-24
feat(ui): add breadcrumb on react pages

kduret:
## Description

Add breadcrumb on react pages only
Legacy breadcrumb is still used on legacy pages (iframes)

image

Refs: MON-4049

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [x] 19.04.x
- [x] 19.10.x (master)


www/front_src/src/redux/selectors/navigation.js
```diff
@@ -0,0 +1,129 @@
+import { createSelector } from 'reselect';
+
+// loop on each child and get first available url
+function getFirstAvailableUrl(item) {
```
bdauria: I feel like we could make the flow of this function easier to follow if we divide into sub-functions.

What about:
```suggestion
function findFirstUrl(item) {
  if (item.groups) {
    const groupWithUrl = item.groups.find(findFirstUrl);

    return groupWithUrl ? findFirstUrlInChildren(groupWithUrl) : undefined;
  }

  return item.children ?  findFirstUrlInChildren(item) : undefined;

}

function findFirstUrlInChildren(item) {
  const childWithUrl = item.children.find((child) => child.url);

  return childWithUrl ? getUrl(childWithUrl) : undefined;
}

function getUrl(item) {
  return item.is_react ? item.url : `/main.php?p=${item.page}${item.options}`; 
}
```
It would be great to add some unit tests for that as well

www/front_src/src/components/breadcrumbWrapper/index.jsx
```diff
@@ -0,0 +1,36 @@
...
+  return (
+    <>
+      <Breadcrumb breadcrumbs={breadcrumbPath}/>
+      {children}
+    </>
+  );
+}
+
+const mapStateToProps = (state) => ({
+  breadcrumbs: breadcrumbsSelector(state),
```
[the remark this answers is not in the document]
kduret: I can't until #7724 is merged

www/front_src/src/redux/selectors/navigation/breadcrumbs.js
```diff
@@ -0,0 +1,106 @@
...
+
+// loop on each group/child to get first url
+function findFirstUrl(item) {
+  if (item.groups) {
+    const groupWithUrl = item.groups.find(findFirstUrl);
+
+    return groupWithUrl ? findFirstUrlInChildren(groupWithUrl) : undefined;
+  }
+
+  return item.children ?  findFirstUrlInChildren(item) : undefined;
```
jdelpierre: little comment : 
2 whitespaces between "?" & "findFirstUrl...

PR #7748, opened 2019-08-01
Remove deprecated engine fields

kduret:
Remove useless engine configuration parameters : 
* check_result_path
* use_check_result_path
* max_check_result_file_age


lpinsivy:
Check with MAP to have synchronization for 19.10

kduret:
we don't merge it right now cause it may impact users which use mod_gearman or MKLiveStatus

ganoze:
@kduret @lpinsivy I guess that this PR **can** make it to the 19.10 release because it is unlikely that Centreon Engine has any compatibility left with either MKLiveStatus or mod_gearman after its heavy refactoring.

PR #7784, opened 2019-08-19
Delete services when host template is detached from host

miteto:
# Pull Request Template
Delete services when host template is detached from host

## Description
Include DB-Func file and use deleteHostServiceMultiTemplate() to delete related services for host templates.

## Type of change

## Target serie

- [ ] 2.8.x
- [X] 18.10.x
- [X] 19.04.x
- [X] 19.10.x (master)

### DIFF
--- a/lib/Centreon/Object/Relation/Host/Template/Host.php
+++ b/lib/Centreon/Object/Relation/Host/Template/Host.php
@@ -71,6 +71,30 @@ class Centreon_Object_Relation_Host_Template_Host extends Centreon_Object_Relati
     }
 
     /**
+     * Delete host template / host relation and linked services
+     *
+     * @param int $fkey - host template
+     * @param int $skey - host
+     * @return void
+     */
+    public function delete($fkey, $skey = null): void
+    {
+        global $pearDB;
+        $this->db->beginTransaction();
+        try {
+            parent::delete($fkey, $skey);
+            $pearDB = $this->db;
+            $centreon = true; // Needed so we can include file below
+            require_once __DIR__."/../../../../../../www/include/configuration/configObject/host/DB-Func.php";
+            deleteHostServiceMultiTemplate($skey, $fkey, array(), null);
+            $this->db->commit();
+        } catch (\PDOException $e) {
+            $this->db->rollBack();
+            exitProcess(PROCESS_ID, 1, $e->getMessage());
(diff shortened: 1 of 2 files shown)


lib/Centreon/Object/Relation/Host/Template/Host.php
```diff
@@ -70,6 +70,30 @@ public function insert($fkey, $skey = null)
...
+     */
+    public function delete($fkey, $skey = null): void
+    {
+        global $pearDB;
+        $this->db->beginTransaction();
+        try {
+            parent::delete($fkey, $skey);
+            $pearDB = $this->db;
+            $centreon = true; // Needed so we can include file below
+            require_once __DIR__."/../../../../../../www/include/configuration/configObject/host/DB-Func.php";
```
vhr: maybe will be better if you replace `__DIR__."/../../../../../../` with `_CENTREON_PATH_` constant, how do you think?

sc979: +1

sc979: style : please add spaces around the dot

lib/Centreon/Object/Relation/Host/Template/Host.php
```diff
@@ -70,6 +70,30 @@ public function insert($fkey, $skey = null)
...
+     * @return void
+     */
+    public function delete($fkey, $skey = null): void
+    {
+        global $pearDB;
+        $this->db->beginTransaction();
+        try {
+            parent::delete($fkey, $skey);
+            $pearDB = $this->db;
+            $centreon = true; // Needed so we can include file below
```
sc979: why do you remove the global $centreon in DB-FUNC if you need it ?

miteto: You can see here :
https://github.com/centreon/centreon/blob/19.04.x/www/include/configuration/configObject/host/DB-Func.php#L36
that $centreon needs to be set, otherwise it will exit().
In the method itself - I removed it because it is not used anywhere.

sc979: that's why I don't understand you want affect a boolean instead of this : 
image
why not getting the $centreon using global ?

PR #7934, opened 2019-10-08
fix(sql): update column types of downtimes table

kduret:
## Description

broker could not insert max timestamp as end_time and duration.
end_time now allow timestamp corresponding to max of 2*64

Fixes BAM-740

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [x] 19.10.x (master)


sc979:
are you sure, that this modification is not needed in previous major versions ?

kduret:
not needed cause it's related to strict mode ;)

PR #7944, opened 2019-10-09
fix(remote-server): fix import with strict mode

kduret:
## Description

remote server import does not work with strict mode enabled
TRUNCATE TABLE does not work on tables with TEXT or BLOB columns

**Fixes**MON-4230

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


src/CentreonRemote/Domain/Exporter/ConfigurationExporter.php
```diff
@@ -80,8 +80,10 @@ public function import(ExportManifest $manifest): void
 
             $import = $manifest->get("import");
             foreach ($import['data'] as $data) {
-                // truncate table
-                $db->query("TRUNCATE TABLE `" . $data['table'] . "`;");
+                // empty table
+                $db->query("DELETE FROM `" . $data['table'] . "`");
+                // optimize table
+                $db->query("ALTER TABLE `" . $data['table'] . "`");
```
sc979: will $data['table'] be able to contain something like "MODIFY", "ADD COLUMN" or some other like queries needing a "IF NOT EXIST" if being already added to the table ?
If so, it won't work on deported MySQL 8 DB

ganoze: Can you expected the expected side effect of this query ? Why not use a clearer query like `OPTIMIZE TABLE` or `ANALYSE TABLE` ?

kduret: I've updated to OPTIMIZE TABLE
I thought ALTER TABLE rebuild data and index too. But regarding documentation, it does not..

Issue #7306, opened 2019-03-20
Problem Reload Service Filter


ptitpopey:
<h1>Problem Reload Service Filter </h1>

***Version 2.8.27***

When you change filter "Status service" in Supervision -> Status Details, the page dosen't refresh. You need to wait refresh time to see the status of service

ptitpopey:
This problem appear only 2.8.27 version. I have downgrade to 2.8.26 and the problem disapear.

CPbN:
Same issue here.

Go to Monitoring / Services.
Service status is by default set to "Unhandled Problems".
Set it to All.

It will automatically update 15 seconds later, but not immediately.
If you want to update it immediately, you have to toggle it to another value first...

CPbN:
Same issue here.

Go to Monitoring / Services.
Service status is by default set to "Unhandled Problems".
Set it to "All".

It will automatically update 15 seconds later, but not immediately.
If you want to update it immediately, you have to toggle it to another value first...

Edit : issue still present in 19.04.0.

BenoitPoulet:
Hello,

We upgrade to 2.8.27, and same pb.

This is really painful, many users are complaining about this. Please provide a fix.

CPbN:
Culprits are here @callapa :) :
(2.8.x) https://github.com/centreon/centreon/commit/f41e5408e4eb8ddcfa376808574b90b4e850a860
(18.10.2+) https://github.com/centreon/centreon/commit/1bf7ff5276bc1415d4e8eb8762bf555554d9bf0f

Where the following has been removed from `service.php` :
```
window.clearTimeout(_timeoutID);
initM(_tm, _sid, _o);
```

Unfortunately `change_type_order()` which is called in `updateSelect()` does not refresh monitoring list if order remains the same...

lpinsivy:
The Centreon web 2.8.28 has been released

CPbN:
@lpinsivy I did not test 19.05.0 yet, but issue was still present in 19.04.0...

CPbN:
Also fixed in 19.04.1 :+1:  Thx !

ptitpopey:
Issue #7306 not resolved. I have upgrade to 2.8.28 and the problem appear again.

ptitpopey:
Please re-open this issue

CPbN:
[...]

PR #7961, opened 2019-10-11
Fix status details not reloading when filter changed, #7306

sbraz:
The problem was introduced by f41e5408e4eb8ddcfa376808574b90b4e850a860

### DIFF
--- a/www/include/monitoring/status/Services/service.php
+++ b/www/include/monitoring/status/Services/service.php
@@ -543,5 +543,7 @@ $tpl->display("service.ihtml");
     function statusServices(value, isInit)
     {
         _o = value;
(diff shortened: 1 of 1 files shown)


sbraz:
@kduret @sc979 is there any chance you guys could merge this?

sc979:
Hi @sbraz ,

Thanks again for your contribution and sorry for the delay in our reply.

Unfortunately after dicussions, the Team decided that the PR won't be merged, as on the 2.8.x version, only security fixes are released.
Furthermore, we removed (for security issues) the use of the _sid.

For those reasons, I close this PR.

Regards.
sc979

PR #8061, opened 2019-10-25
Add function to try get hostname by client mysql

lgcosta:
# Pull Request Template

## Description

Add function `gethostbyaddr` to try get the hostname of IP address by connection of mysql client.

For example, Docker and Docker-compose changes the IP in each restart and it may happen that ip is not the same as configured in mysql authorization table. That way, if you register the IP hostname, you will have no problem with changing IP.

Ref: https://www.php.net/manual/en/function.gethostbyaddr.php

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [x] 19.10.x (master)

- [ ] I followed the **coding style guidelines** provided by Centreon
- [ ] I have **rebased** my development branch on the base branch (master, maintenance).

#### Centreon team only

- [ ] I have made sure that the **unit tests** related to the story are successful.
- [ ] I have made sure that **unit tests cover 80%** of the code written for the story.
- [ ] I have made sure that **acceptance tests** related to the story are successful (**local and CI**)

### DIFF
--- a/www/install/steps/process/createDbUser.php
+++ b/www/install/steps/process/createDbUser.php
@@ -66,7 +66,7 @@ if ($parameters['address'] != "127.0.0.1" && $parameters['address'] != "localhos
     );
     $getIpQuery->execute();
     // The result example (172.17.0.1:38216), use the explode function to remove port
-    $host = explode(":", $getIpQuery->fetchAll(PDO::FETCH_COLUMN)[0])[0];
+    $host = gethostbyaddr(explode(":", $getIpQuery->fetchAll(PDO::FETCH_COLUMN)[0])[0]);
 }
 
 // Compatibility adaptation for mysql 8 with php7.1 before 7.1.16, or php7.2 before 7.2.4.


lgcosta:
> I am not comfortable with using _gethostbyaddr()_ which probably translates to a reverse DNS lookup on the client machine (not on the server). We might end up with random connection issues that will be quite hard to diagnose.
> 
> I am unsure on how we expect the MySQL client address filtering to work. Should it be by IP address or by host name ? Should it be automatic or configurable ?

This function will always use dns server configured on the PHP server, ie whatever is set to `/etc/resolve.conf` or  get via dhcp.  The idea is to get a hostname from reverse resolution configured for that IP on the dns server, if that fails it will simply return the same IP.

Capturing this session is already automated from the active session at query time, just thinking of a complex environment that might involve for example a proxy.  This automation ensures that the captured ip is the same one used in the active session (if it can query it is because it is valid).

If there is a reverse hostname configured for the IP and is incorrect, not only this function has problems, such as other system applications.

lgcosta:
In example with docker-compose, I can force a network configuration into the environment and make the container always have a fixed IP, but I would do this because I know this happens, but we can still have other environments that can suffer IP changes without a customization of that level.

lpinsivy:
Hi @lgcosta we think this can have a side effect.

It could be better to get IP of docker using a script (docker inspect)

lgcosta:
> Hi @lgcosta we think this can have a side effect.
> 
> It could be better to get IP of docker using a script (docker inspect)

This is a bit complicated because it has to be done within an instance of the docker. But it is nothing "out of this world". No problem ;-)

PR #8146, opened 2019-11-25
fix(acl): fix calculation of acls on services

kduret:
## Description

ACLs on service were not calculated (only on 19.04.5), so non-admin users could not see any services.

Refs: MON-4530

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [x] 19.04.x
- [ ] 19.10.x
- [ ] 20.04.x (master)

### DIFF
--- a/cron/centAcl-Func.php
+++ b/cron/centAcl-Func.php
@@ -125,39 +125,38 @@ function getCentAclRunningState()
  * Return host tab after poller filter
  *
  * @param array $host
- * @param integer $res_id
+ * @param integer $resId
  * @return array $host
  */
-function getFilteredPollers($host, $res_id)
+function getFilteredPollers($host, $resId)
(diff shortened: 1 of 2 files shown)


cron/centAcl-Func.php
```diff
@@ -534,14 +531,14 @@ function getMyHostServicesByName($host_id = null)
  */
 function getMetaServices($resId, $db, $metaObj)
 {
-    $sql = "SELECT meta_id FROM acl_resources_meta_relations WHERE acl_res_id = {$db->escape($resId)}";
+    $sql = "SELECT meta_id FROM acl_resources_meta_relations WHERE acl_res_id = " . (int) $resId;
```
sc979: why not prepare the query here ?

PR #8286, opened 2020-01-24
fix(graph): manage timezone on zoom

kduret:
## Description

Timezone is not kept when zoom is applied on a graph

**Fixes** MON-4510

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [x] 19.04.x
- [x] 19.10.x
- [x] 20.04.x (master)

* update timezone of logged user (eg: Asia/Tokyo)
* zoom on a chart and check that x axis is updated properly following timezone


www/include/views/graphs/graphs.html
```diff
@@ -129,6 +129,12 @@
 var templateGraph = Handlebars.compile(source);
 var autoRefresh = false;
 
+// get timezone from localstorage
+// be careful, it seems that it's updated when user logout/login
+const timezone = localStorage.getItem('realTimezone')
+    ? localStorage.getItem('realTimezone')
+    : moment.tz.guess();
```
sc979: this way you'll get the timezone from the browser. But if the user change this value in the browser, will it be reflected without the need of closing/reopening centreon's tab ?
I mean, shouldn't we use the "ignore cache" option of moment, like in the example from the doc : 
```
moment.tz.guess(); // America/Chicago
// suppose the client's timezone changes to Europe/Berlin
moment.tz.guess(); // America/Chicago
moment.tz.guess(true); // Europe/Berlin
```

kduret: We don't have a single page application, so I guess it does not change anything to use cache or not for this :)
But anyway, I updated source code to use ignoreCache parameter

PR #8353, opened 2020-02-20
doc(api): update unified view contract

kduret:
## Description

add some missing fields in API contract of unified view

## Type of change

- [x] Updating documentation (missing information, typo...)

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


doc/API/centreon-api-v2.yaml
```diff
@@ -1646,7 +1676,7 @@ components:
                 url:
                   type: string
                   description: "Url of the icon"
-                  example: "/media/linux.png"
+                  example: "./img/media/topology/firewall.png"
```
bdauria: Is the path going to be relative?

kduret: yes. Is it better to manipulate an absolute path on frontend ?

kduret: final decision: use absolute path and frontend will add root path (eg: /centreon)

PR #8381, opened 2020-02-28
feat(api): add eventlog listing

victorvassilev:
## Description

Add endpoint to allow retrieving event timeline for hosts and services

**Fixes** # MON-4866,MON-4867

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

Visit the endpoint using GET:

http://host/centreon/api/beta/monitoring/hosts/14/services/19/timeline
http://host/centreon/api/beta/monitoring/hosts/14/timeline

- [x] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [x] I have commented my code, especially **hard-to-understand areas** of the PR.
- [x] I have made corresponding changes to the **documentation**.

#### Centreon team only

- [x] I have made sure that the **unit tests** related to the story are successful.
- [ ] I have made sure that **unit tests cover 80%** of the code written for the story.
- [x] I have made sure that **acceptance tests** related to the story are successful (**local and CI**)


src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
```diff
@@ -0,0 +1,430 @@
...
+        $statement->execute();
+
+        $result = $this->db->query('SELECT FOUND_ROWS()');
+        $this->sqlRequestTranslator->getRequestParameters()->setTotal(
+            (int)$result->fetchColumn()
+        );
+
+        while (false !== ($result = $statement->fetch(\PDO::FETCH_ASSOC))) {
+            $newEvent = EntityCreator::createEntityByArray(
+                $this->detectClass($result['eventType']),
```
kduret: I think this could be handled by a serializer discriminator. But let's keep it like that for now

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
```diff
@@ -0,0 +1,430 @@
...
+     *
+     * @param \Centreon\Infrastructure\CentreonLegacyDB\StatementCollector $collector
+     * @param int $hostId
+     * @param int $serviceId
+     * @return string
+     */
+    protected function generateLogsQuery(StatementCollector $collector, int $hostId, int $serviceId = null): string
+    {
+        $sql = "SELECT
+		CONCAT('L', l.log_id) AS `eventId`,
```
kduret: IMHO, eventId should be an integer from only eventId
The unique id will be the association between eventType and eventId

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
```diff
@@ -0,0 +1,430 @@
...
+     *
+     * @param \Centreon\Infrastructure\CentreonLegacyDB\StatementCollector $collector
+     * @param int $hostId
+     * @param int $serviceId
+     * @return string
+     */
+    protected function generateCommentsQuery(StatementCollector $collector, int $hostId, int $serviceId = null): string
+    {
+        $sql = "SELECT
+		CONCAT('C', c.comment_id) AS `eventId`,
```
kduret: IMHO, eventId should be an integer from only eventId
The unique id will be the association between eventType and eventId

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
```diff
@@ -0,0 +1,430 @@
...
+     *
+     * @param \Centreon\Infrastructure\CentreonLegacyDB\StatementCollector $collector
+     * @param int $hostId
+     * @param int $serviceId
+     * @return string
+     */
+    protected function generateDowntimesQuery(StatementCollector $collector, int $hostId, int $serviceId = null): string
+    {
+        $sql = "SELECT
+		CONCAT('D', d.downtime_id) AS `eventId`,
```
kduret: IMHO, eventId should be an integer from only eventId
The unique id will be the association between eventType and eventId

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
```diff
@@ -0,0 +1,430 @@
...
+     * @return string
+     */
+    protected function generateAcknowledgementsQuery(
+        StatementCollector $collector,
+        int $hostId,
+        int $serviceId = null
+    ): string {
+    
+        $sql = "SELECT
+		CONCAT('A', a.acknowledgement_id) AS `eventId`,
```
kduret: IMHO, eventId should be an integer from only eventId
The unique id will be the association between eventType and eventId

src/Centreon/Domain/Monitoring/Entity/CommentEventObject.php
```diff
@@ -0,0 +1,79 @@
...
+     */
+    public function getEventId(): string
+    {
+        return self::EVENTTYPE . $this->getId();
+    }
+
+    /**
+     * {@inheritdoc}
+     */
+    public static function loadEntityDescriptorMetadata(): array
```
callapa: What is the purpose of this method ?

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
victorvassilev: no it can't because the discriminator requires a specific field of the entity to detect which instance it is, in this case we have a class that has a subobject which changes implementations, I tried to do it with discriminator and it didnt work well

src/Centreon/Infrastructure/Monitoring/TimelineRepositoryRDB.php
victorvassilev: this way it wouldn't be unique :) we just needed 1 field that is unique for every timeline set of data

kduret: We can do a unique identifier based on two fields
For instance, if we want to get information of a specific event, the controller will look like:
`monitoring/events/{type}/{id}`

victorvassilev: I agree with you, but this 'eventId' property does not need to be used if FE don't need it, and can be removed later if we decide it is useless. For now it does not hurt to have it, because I believe every listing endpoint had to have unique identifier for each row of returned info (due to FE requirements). If that is not the case I can always just omit it from the return of the endpoint. It does not create extra load, it does not mess with any business logic, so in a future refactor we can just remove it easily.

PR #8393, opened 2020-03-05
feat(doc): update the API documentation

vhr:
## Description

Extend the Resource list endpoint with query parameters for each drop-down in the filter bar

**Fixes** MON-4887


doc/API/centreon-api-v2.yaml
```diff
@@ -183,13 +183,46 @@ paths:
...
+        * parent_status
+        * severity_level
+        * in_downtime
+        * acknowledged
+        * impacted_resources_count
+        * last_status_change
+        * tries
+        * last_check
+        * information
+        * host.group
```
kduret: not sure we'll allow to order by host group

vhr: why we have to restrict if it's possible?

kduret: because, for instance, a BA is not linked to a hostgroup

PR #8426, opened 2020-03-19
enh(ui): improve styling of unified view filters

kduret:
## Description

- add collapse
- increase width of global search
- put "filter" & "criterias" labels inlint

image

**Fixes** MON-5014

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


package.json
```diff
@@ -67,12 +69,13 @@
     "webpack-merge": "^4.2.2"
   },
   "dependencies": {
-    "@centreon/ui": "centreon/centreon-ui",
+    "@centreon/ui": "github:centreon/centreon-ui",
```
bdauria: why this change?

www/front_src/src/Resources/Filter/index.tsx
```diff
@@ -40,16 +53,38 @@ import {
...
+    transform: 'rotate(0deg)',
+    marginLeft: 'auto',
+    padding: theme.spacing(1),
+    transition: theme.transitions.create('transform', {
+      duration: theme.transitions.duration.shortest,
+    }),
+  },
+  expandOpenButton: {
+    transform: 'rotate(180deg)',
+  },
```
bdauria: I think we can live without the animation for now

www/front_src/src/Resources/index.test.tsx
```diff
@@ -163,6 +172,20 @@ describe(Resources, () => {
...
   });
 
+  it('expands criterias filters', async () => {
+    const { queryByText } = render(<Resources />);
+
+    await wait(() => {
+      expect(queryByText(labelTypeOfResource)).not.toBeVisible();
+    });
+
+    expandCriteriasFilters();
```
bdauria: can we use `getByLabelText` instead?

package.json
kduret: it's automatically added since I've upgraded npm

www/front_src/src/Resources/Filter/index.tsx
kduret: Do you want I set transition duration to 0 ?

kduret: Done. I used ExpansionPanel component from material-ui

PR #8474, opened 2020-03-26
feat(ui): allow to schedule check in events view

kduret:
## Description

Allo to schedule check in events view : 
- directly from the line
- by bulk using checkboxes and action button

**Fixes** MON-4924

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

Try to schedule check from events view.


www/front_src/src/Resources/Actions/Acknowledge/index.tsx
```diff
@@ -101,4 +101,8 @@ const AcknowledgeForm = ({
   );
 };
 
-export default AcknowledgeForm;
+export default React.memo(
```
bdauria: We should avoid rendering the Form at all when there is nothing to be displayed instead

www/front_src/src/Resources/Actions/Downtime/index.tsx
```diff
@@ -194,4 +194,8 @@ const DowntimeForm = ({
   );
 };
 
-export default DowntimeForm;
+export default React.memo(
```
bdauria: Same here. Maybe try to re-use the implemented behavior

PR #8478, opened 2020-03-26
fix(listing): set events view listing to empty on api error

kduret:
## Description

when http code 500 is returned from resources endpoint, listing is reset to empty.
Furthermore, the snackbar error message is now more clearer

**Fixes** MON-5034

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

set search filter to `*test`
==> the listing should be empty

### DIFF
--- a/www/front_src/src/Resources/index.tsx
+++ b/www/front_src/src/Resources/index.tsx
@@ -98,7 +98,7 @@ const Resources = (): JSX.Element => {
       })
       .catch((error) => {
         setListing(undefined);
-        showError(error?.response?.data?.message || error.message);
+        showError(error.response?.data?.message || error.message);
       })
       .finally(() => setLoading(false));
   };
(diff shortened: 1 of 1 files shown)


www/front_src/src/Resources/index.tsx
```diff
@@ -97,7 +97,8 @@ const Resources = (): JSX.Element => {
         setListing(retrievedListing);
       })
       .catch((error) => {
-        showError(error.message);
+        setListing(undefined);
+        showError(error?.response?.data?.message || error.message);
```
bdauria: We need to have a consistent way of displaying error messages because right now the behavior is different everywhere.

kduret: as we discussed, the error format sent by backend will be probably like that : 
`{"code":500,"message":"Error while searching for resources"}`

PR #8487, opened 2020-03-28
enh(ui): update style of events view page

kduret:
## Description

Listing is more dense : 
image

**Fixes** MON-5016

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

check new style


package.json
```diff
@@ -72,7 +72,7 @@
     "webpack-merge": "^4.2.2"
   },
   "dependencies": {
-    "@centreon/ui": "github:centreon/centreon-ui#adbb81c3b4dc6bf67842b890c2ac4e175562decd",
+    "@centreon/ui": "github:centreon/centreon-ui#MON-5016-style",
```
bdauria: We need to be careful to merge the corresponding centreon-ui branch before merging this one to avoid breaking the build afterwards

sc979:
The screenshot seems much more user friendly

www/front_src/src/Resources/columns/Graph/index.tsx
```diff
@@ -25,7 +25,7 @@ import { ColumnProps } from '..';
 const JSXXAxis = (XAxis as unknown) as (props) => JSX.Element;
 const JSXYAxis = (YAxis as unknown) as (props) => JSX.Element;
 
-const graphHeight = 350;
+const graphHeight = 200;
```
bdauria: 200 is too small as a height for the graph, the legend can overlap if there are too many elements:

Screenshot from 2020-03-29 11-27-56

kduret: I think the legend can overlap also with 350 height (eg: monitoring of network interfaces of a switch)
It should be better to make a custom legend to manage lot of metrics

kduret: Anyway, I've increased the height to 250

kduret:
> @kduret the padding is too small in two areas: between the action icons and within the table rows. I think it's better to add some more padding (at least one unit), and reduce the size of the icons instead (Status & State chips)

I move cell padding to 0.5 (instead of 0.25) cause 1 was too big, even if I reduce the size of the icons.
I've also added some padding between actions icons

PR #8536, opened 2020-04-03
fix(ui): fix pagination and autocomplete search

kduret:
## Description

- avoid to clear autocomplete search when listing is refreshed
- fix endpoint query when autocomplete search is cleared
- go to page 1 when search or sort is updated

**Fixes** MON-5099

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


www/front_src/src/Resources/Filter/index.tsx
```diff
@@ -142,7 +142,7 @@ const Filter = ({
 }: Props): JSX.Element => {
   const classes = useStyles();
 
-  const [searchFieldValue, setSearchFieldValue] = useState<string>();
+  const [searchFieldValue, setSearchFieldValue] = React.useState<string>();
```
jdelpierre, line 145: Why it's necessary to specify "React" here and not on "import" ?

kduret: Because the recommanded import syntax is : `import * as React from 'react;`

PR #8557, opened 2020-04-07
fix(ui): use timezone in downtime popin

kduret:
## Description

Take into account user timezone in downtime popin

/!\ this does not fix the table columns which do not use timezone

note : did not find a way to unit test timezone in popin

**Fixes** MON-5122

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

- Set a timezone to the user
- Add a downtime to a resource and check that dates and times take into account user timezone

### DIFF
--- a/www/front_src/src/Resources/Actions/Resource/Downtime/Dialog.tsx
+++ b/www/front_src/src/Resources/Actions/Resource/Downtime/Dialog.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 
+import moment from 'moment-timezone/builds/moment-timezone-with-data-10-year-range';
 import MomentUtils from '@date-io/moment';
 
 import { Typography, Checkbox, FormHelperText, Grid } from '@material-ui/core';
@@ -39,6 +40,7 @@ import { Resource } from '../../../models';
 
 interface Props {
   locale: string | null;
+  timezone: string | null;
   resources: Array<Resource>;
   canConfirm: boolean;
   onCancel;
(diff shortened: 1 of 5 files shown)


www/front_src/src/Resources/Actions/Resource/Downtime/Dialog.tsx
```diff
@@ -94,6 +97,14 @@ const DialogDowntime = ({
...
   };
 
+  React.useEffect(() => {
+    moment.locale(locale);
+  }, [locale]);
+
+  React.useEffect(() => {
+    moment.tz.setDefault(timezone);
+  }, [timezone]);
+
```
bdauria, line 107: I think these effects are a bit superfluous. I think we should load the dialog until all data needed are available (including locale and timezone).

kduret: I could put it in a single useEffect with "loading" as dependency, but I don't think it's the best way.
So I think I don't understand what is exactly your point ?

bdauria: No that's not what I mean. I mean that in the Dialog, locale and timezone are not supposed to be null, so we should have only one hook call setting them at the beginning. Also the Dialog is not supposed to be usable if they're not available yet

kduret: we'll probably refactor this once user information are stored in global context

PR #8564, opened 2020-04-08
enh(ui): Fixes to details panel

bdauria:
## Description

This PR makes the following fixes:
-  Show flapping card only when flapping is defined / show 'Yes' label when value is true, 'N/A' otherwise
- Fix command line not showing the correct value, and add a copy button. The card is not shown if command line value is not defined
- Reduce the font sizes of cards content

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

### DIFF
--- a/www/front_src/src/Resources/Details/Body/tabs/Details/DetailsCard/cards.tsx
+++ b/www/front_src/src/Resources/Details/Body/tabs/Details/DetailsCard/cards.tsx
@@ -18,6 +18,7 @@ import {
   labelPercentStateChange,
   labelLastNotification,
   labelCurrentNotificationNumber,
+  labelNo,
 } from '../../../../../translatedLabels';
 import { getFormattedDate, getFormattedTime } from '../../../../../dateTime';
 import { ResourceDetails } from '../../../../models';
@@ -154,7 +155,7 @@ const getDetailCardLines = (
       getLines: (): Lines => [
         {
           key: 'flapping',
-          line: <DetailsLine line={details.flapping ? labelYes : 'N/A'} />,
+          line: <DetailsLine line={details.flapping ? labelYes : labelNo} />,
         },
       ],
     },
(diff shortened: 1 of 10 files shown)


www/front_src/src/Resources/Details/Body/tabs/Details/DetailsCard/cards.tsx
```diff
@@ -148,7 +154,7 @@ const getDetailCardLines = (
       getLines: (): Lines => [
         {
           key: 'flapping',
-          line: <DetailsLine line={details.flapping ? 'N/A' : labelYes} />,
+          line: <DetailsLine line={details.flapping ? labelYes : 'N/A'} />,
```
kduret, line 157: if flapping is false, it will display "N/A" ?

bdauria: Yes, the rule is -> defined ? show the card ? truish? display 'Yes', 'N/A' otherwise

kduret: IMHO, it should always display the card with this rule : 
- not defined : N/A
- defined and true : yes
- defined and false : no

(currently, flapping seems not returned by API...)
we may need to check with @Moujimouja what is exactly the behaviour we want

bdauria: @kduret it is actually returned in some cases.

kduret: there is no any information related to flapping in backend serializer : https://github.com/centreon/centreon/blob/master/config/packages/serializer/Centreon/Monitoring.Resource.yml
This information should be returned in any cases

PR #8578, opened 2020-04-09
enh(ui): Add delay to HoverChip tooltips and allow display of all results in DetailsTable

bdauria:
## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

### DIFF
--- a/www/front_src/src/Resources/columns/State/DetailsTable/index.tsx
+++ b/www/front_src/src/Resources/columns/State/DetailsTable/index.tsx
@@ -36,7 +36,7 @@ const DetailsTable = <TDetails extends unknown>({
   endpoint,
   columns,
 }: DetailsTableProps): JSX.Element => {
-  const [details, setDetails] = useState<TDetails | null>();
+  const [details, setDetails] = useState<Array<TDetails> | null>();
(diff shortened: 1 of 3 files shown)


www/front_src/src/Resources/columns/State/DetailsTable/index.tsx
```diff
@@ -69,24 +69,30 @@ const DetailsTable = <TDetails extends unknown>({
...
-              columns.map(({ label, getFormattedString }) => (
-                <TableCell key={label}>
-                  <span>{getFormattedString?.(details)}</span>
-                </TableCell>
-              ))}
-            {error && (
+            </TableRow>
+          )}
+          {success &&
+            details?.map((detail) => (
```
kduret, line 80: details could be initialize to [] to avoid optional chaining

bdauria: Right now it wouldn't work, because 'null' is the error case and is a potential value; this makes TypeScript prevent us from calling map on details without error.

I'll refactor the DetailsTable in another PR to clean things up.

PR #8580, opened 2020-04-10
enh(ui): improve search help tooltip in events view

kduret:
## Description

Improve search help tooltip in events view : 
- stay persistent, until the close button is clicked
- add close icon button
- set background to white
- some minor improvements to avoid unneccessary rendering (remove arrow functions)

image

**Fixes** MON-5211

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

Check help tooltip in events view

### DIFF
--- a/www/front_src/src/Resources/index.test.tsx
+++ b/www/front_src/src/Resources/index.test.tsx
@@ -28,6 +28,8 @@ import { ThemeProvider } from '@centreon/ui';
 import {
   labelResourceName,
   labelSearch,
+  labelSearchHelp,
+  labelSearchOnFields,
   labelInDowntime,
   labelAcknowledged,
   labelTypeOfResource,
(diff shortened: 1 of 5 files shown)


www/front_src/src/Resources/index.test.tsx
```diff
@@ -1054,4 +1058,28 @@ describe(Resources, () => {
       }),
     );
   });
+
+  it('leaves search help tooltip when the search input filled ', async () => {
```
bdauria, line 1062: What are we testing here exactly? that the search input is still visible even if the help tooltip is displayed?

kduret: that the tooltip is not closed when searching.
Cause it will be the main use case for user who want to get this help when typing a search

PR #8585, opened 2020-04-12
enh(ui): Improve Graph display when no data are available and legend is large / fix clear all command

bdauria:
## Description
This PR improves the display of Graph, when either the legend is very large, and no data is available. The layout has been made more responsive so that it adapts better to both the tooltip and the details panel. 

Example with different kinds of legend data:
Screenshot from 2020-04-12 19-15-01
Screenshot from 2020-04-12 19-18-44
Screenshot from 2020-04-12 19-19-18
Screenshot from 2020-04-12 19-20-29
Screenshot from 2020-04-12 20-37-32

It also fixes the "clear all" so that it also clears host groups, service groups and the search field. 

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


www/front_src/src/Resources/Details/Body/tabs/Details/index.tsx
```diff
@@ -135,39 +140,31 @@ const DetailsTab = ({ details }: Props): JSX.Element => {
...
-                  </Grid>
+        <Card>
+          <CardContent>
+            <Typography variant="subtitle2" color="textSecondary" gutterBottom>
+              <Grid container alignItems="center" spacing={1}>
+                <Grid item>{labelCommand}</Grid>
+                <Grid item>
+                  <Tooltip onClick={copyCommandLine} title={labelCopy}>
+                    <IconButton size="small">
+                      <IconCopyFile color="primary" fontSize="small" />
```
kduret, line 157: We'll may need to remove the primary color. Cause I think the icon is too visible regarding important information of the panel

kduret: But we'll wait feedbacks to manage this point

jdelpierre:
On the second screenshot, it’s note possible to align bullets?

bdauria:
> Why the first screenshot displayed in PR description does not display "No data available for this period" ?

Because the API dosen't return an empty `times`

bdauria:
> On the second screenshot, it’s note possible to align bullets?

It would be possible the use a centered justification instead of space-between like we do, however we wouldn't handle the case with a lot of metrics displaying large legend texts properly

kduret:
> > Why the first screenshot displayed in PR description does not display "No data available for this period" ?
> 
> Because the API dosen't return an empty `times`

Ok, Then we'll probably investigate in the backend part

kduret:
But it's weird, the details panel is for service `Ping`, but the graph is for `proc-ntpd`

bdauria:
> But it's weird, the details panel is for service `Ping`, but the graph is for `proc-ntpd`

Yes, I used real data but mocked out the endpoint to be able to create these screenshots, so the charts don't correspond to the resources.

bdauria:
@jdelpierre I used a new approach to be able to justify legend items in a centered way. The drawback is that the graph is now globally scrollable within the tooltip if a lot of metrics are displayed, which is more acceptable than the other option in my opinion.

kduret:
css looks much more understandable now :)

PR #8586, opened 2020-04-13
fix(ui): use last value of search field when a filter triggers a listing search

kduret:
## Description

Use last value of search field when a filter triggers a listing search

**Fixes** MON-5228

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

- Type something in events view search field without clicking on search button (and without pressing enter key)
- Change another filter (eg: "All" instead of "Unhandled Problems"
- Check the listing take into account search field

### DIFF
--- a/www/front_src/src/Resources/index.tsx
+++ b/www/front_src/src/Resources/index.tsx
@@ -163,6 +163,17 @@ const Resources = (): JSX.Element => {
     load();
   };
 
+  const prepareSearch = (event): void => {
+    setNextSearch(event.target.value);
+  };
+
+  const requestSearch = (): void => {
+    if (currentSearch === nextSearch) {
+      initAutorefreshAndLoad();
+    }
+    setCurrentSearch(nextSearch);
+  };
+
   React.useEffect(() => {
     return (): void => {
       tokenSource.cancel();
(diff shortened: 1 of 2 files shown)


www/front_src/src/Resources/index.tsx
```diff
@@ -172,34 +183,22 @@ const Resources = (): JSX.Element => {
...
-    hostGroups,
-    serviceGroups,
-  ]);
+  }, [sortf, sorto, page, limit, currentSearch]);
+
+  const firstUpdate = React.useRef(true);
+  React.useEffect(() => {
+    if (firstUpdate.current) {
+      // to avoid duplicate rendering on mount
+      firstUpdate.current = false;
```
bdauria, line 192: What causes the double render?

kduret: There are 2 useEffect which induces double render : 

- https://github.com/centreon/centreon/pull/8586/files#diff-ca62ddbe4609c0067343bddbddf3f09aR184
- https://github.com/centreon/centreon/pull/8586/files#diff-ca62ddbe4609c0067343bddbddf3f09aR189

kduret: the other solution is to use `setCurrentSearch(nextSearch);` in each function who handle filter changes, but I think it's too verbose

PR #8628, opened 2020-04-18
chore(dev): Setup webpack-dev-server with HMR

bdauria:
## Description
This PR introduces the webpack-dev-server with HMR and react-hot-loader. This improves greatly the development workflow by reloading modules on the fly when something changes instead of having to reload the whole page.

HMR is not supported for external modules yet as it's not compatible with System js at the moment. We might need to wait for webpack 5 to be able to build external ES modules or give https://github.com/alexisvincent/systemjs-hot-reloader a try

**This PR should be merged after final 20.04 build**

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [ ] 20.04.x (master)
- [x] 20.10.x (next)

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,6 +48,7 @@
     "eslint-plugin-prettier": "^3.1.2",
     "file-loader": "^6.0.0",
     "fork-ts-checker-webpack-plugin": "^4.1.2",
+    "html-webpack-harddisk-plugin": "^1.0.1",
     "html-webpack-plugin": "^4.0.2",
     "jest": "^25.2.3",
     "jest-environment-jsdom-sixteen": "^1.0.3",
@@ -71,7 +72,6 @@
     "webpack": "^4.42.1",
     "webpack-bundle-analyzer": "^3.6.1",
     "webpack-cli": "^3.3.11",
-    "webpack-livereload-plugin": "^2.3.0",
     "webpack-merge": "^4.2.2"
   },
   "dependencies": {
@@ -92,7 +92,6 @@
     "filesize": "^6.1.0",
     "formik": "^2.1.4",
     "html-react-parser": "^0.10.3",
-    "html-webpack-harddisk-plugin": "^1.0.1",
     "install": "^0.13.0",
     "jsdom": "^16.2.1",
     "loaders.css": "^0.1.2",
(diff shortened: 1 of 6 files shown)


webpack.config.dev.js
```diff
@@ -1,17 +1,29 @@
...
 const path = require('path');
 
 const devConfig = require('@centreon/frontend-core/webpack/patch/dev');
 const baseConfig = require('./webpack.config');
 
+const devServerPort = 9090;
+
 module.exports = merge(baseConfig, devConfig, {
+  output: {
+    publicPath: `http://localhost:${devServerPort}/static/`,
```
kduret, line 11: Does it work if browser is not on the same machine than the build

package.json
```diff
@@ -90,6 +92,7 @@
     "filesize": "^6.1.0",
     "formik": "^2.1.4",
     "html-react-parser": "^0.10.3",
+    "html-webpack-harddisk-plugin": "^1.0.1",
```
kduret, line 95: should be a dev dependency

webpack.config.dev.js
```diff
@@ -1,17 +1,29 @@
...
+  },
   resolve: {
     alias: {
       react: path.resolve('./node_modules/react'),
+      'react-dom': '@hot-loader/react-dom',
       'react-router-dom': path.resolve('./node_modules/react-router-dom'),
       '@material-ui/core': path.resolve('./node_modules/@material-ui/core'),
     },
   },
-  plugins: [new LiveReloadPlugin({ appendScriptTag: true })],
```
kduret, line 16: must be removed from package.json

webpack.config.dev.js
bdauria: It works even if the index.html file is served remotely, but the dev server must run locally.

webpack.config.js
```diff
@@ -20,9 +25,11 @@ module.exports = merge(baseConfig, extractCssConfig, {
   },
   plugins: [
     new HtmlWebpackPlugin({
+      alwaysWriteToDisk: true,
```
jdelpierre, line 28: this is for a better performance... or something else?

bdauria: We need to write the HTML file on disk, not just serve it. It's needed by the docker container which is already serving the file.

Thebarda:
Tested with latest commit and LGTM. 
Build time roughly increased maybe two times less 😍

Issue #6548, opened 2018-08-21
Load the default page when user is already authenticated


UrBnW:
Hi,

The _Default page_ setting works at user login (when user enters his username/password). Perfect.

However, it does not work when the browser is already authenticated, and user navigates to `http://server/centreon`. User is then not redirected and sees the _Home > Custom Views_ page.

Could it be possible to load the _Default page_ here too ?

Thank you very much :+1: 

Edit : issue still present in 19.10.10.

UrBnW:
See #8354 for a fix 👍

kduret:
Fixed with Centreon Web 20.04.1

PR #8354, opened 2020-02-20
fix(UI): redirect user to his default page

UrBnW:
Hi,

## Description

When a user is already logged-in, he will not be redirected to his default page in new tabs / windows.
This PR then redirects the user to his default page reaching Centreon portal, when already logged-in.

It then closes #6548

## Type of change

- [x] Patch fixing an issue (non-breaking change)

## Target serie

- [x] 19.10.x
- [x] 20.04.x (master)

<h2> How this pull request can be tested ? </h2>

Login to Centreon.
Open a new tab or window to your Centreon portal, and be sure to reach your own default page.

Edit : issue still present in 19.10.10.

PR #8644, opened 2020-04-23
fix(ui): fix user default page redirection

kduret:
## Description

Fix default user page when user is already logged in (access to http://<ip_address>/centreon)

**Fixes** MON-5369

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

- Modify user default page
- Log out & Log in
- Go to root page : http://<ip_address>/centreon

This test needs to be done with admin and non admin users

### DIFF
--- a/www/include/common/common-Func.php
+++ b/www/include/common/common-Func.php
@@ -2212,7 +2212,14 @@ function cleanString($str)
 
 // Global Function
 
-function get_my_first_allowed_root_menu($lcaTStr, $defautPage = null)
+/**
+ * get first menu entry allowed to a given user
+ *
+ * @param string $lcaTStr Allowed topology pages separated by comma
+ * @param int $defautPage User default page
+ * @return array The topology information (url, options, name...)
+ */
+function getFirstAllowedMenu($lcaTStr, $defautPage = null)
 {
     global $pearDB;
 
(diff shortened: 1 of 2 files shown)


UrBnW:
Does this PR then fixes #6548 / #8354 ?

www/include/common/common-Func.php
```diff
@@ -2212,31 +2212,38 @@ function cleanString($str)
...
-            . "AND (topology_parent IS NULL OR topology_page = 104) "
-            . "AND topology_page IS NOT NULL AND topology_show = '1' "
-            . "ORDER BY CASE topology_page WHEN 104 THEN 1 ELSE 0 END DESC, topology_id ASC "
-            . "LIMIT 1";
-    } else {
-        $rq = "SELECT topology_parent,topology_name,topology_id,topology_url,topology_page,topology_url_opt, is_react "
-            . "FROM topology "
-            . "WHERE topology_parent IS NULL AND topology_page IS NOT NULL AND topology_show = '1' "
-            . "LIMIT 1";
+    $eventsViewPage = 104;
```
callapa, line 2226: Is it possible to use a constant instead?

kduret:
> Does this PR then fix #6548 / #8354 ?

Yes :)

www/include/common/common-Func.php
kduret: @callapa how would you proceed in that case ? Cause we're not in a class, and I would like to avoid to declare a global constant

PR #8649, opened 2020-04-23
fix(ui): manage timezone in details panel

kduret:
## Description

Manage user timezone in events view details panel

**Fixes** MON-5317

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

* Update user timezone
* Check host / service details display date in the proper timezone


www/front_src/src/Resources/dateTime.ts
```diff
@@ -1,14 +1,13 @@
-import format from 'date-fns/format';
-import parseISO from 'date-fns/parseISO';
+import moment from 'moment';
```
bdauria, line 1: We already embed full moment in the bundle right?

kduret: Yes, and partially moment-timezone

Thebarda:
aaa
It seems to work when opening Downtime dialog

dsdsd
After updating check, it does not update last check date time in the correct timezone

Thebarda:
Tested with last build and it works perfectly

PR #8677, opened 2020-05-01
enh(refacto): Refactor Resources to suppress test warnings

bdauria:
## Description
This is an attempt to prevent the Resources tests from leaking. 

## Type of change

## Target serie
- [x] 20.04.x (master)


www/front_src/src/Resources/index.test.tsx
```diff
@@ -759,8 +730,10 @@ describe(Resources, () => {
 
     const chipLabel = `${acknowledgedEntity?.name} ${labelAcknowledged}`;
 
-    fireEvent.mouseEnter(getByLabelText(chipLabel));
-    fireEvent.mouseOver(getByLabelText(chipLabel));
+    const chip = await findByLabelText(chipLabel);
+
+    fireEvent.mouseEnter(chip);
+    fireEvent.mouseOver(chip);
```
kduret, line 737: performance could be also improved by using `within` on the next line
something like : 
```
const tooltip = getByRole('tooltip');
expect(within(tooltip).getByText('admin')).toBeInTheDocument();
...
```

bdauria:
@kduret I've tried to make different kinds of performance optimizations and we indeed gained some performing time, but we're still hitting timeouts quite regularly on some tests.

I'm trying to think we probably reached the limits of jsdom in terms of timeout with the current architecture, so the only solution that I see so far is that first: we simply set the timeout to 10s - second: we refactor out to split up to different, independently tested Components (Filter, Listing, Dialogs), and try to reduce the timeout.

With that solution, we will see some improvement for sure in test running time, and it'll be easier to debug tests as well **BUT** we will lose the integration validation approach that we maintained so far. We might want to leave that to the future Acceptance framework.

wdyt?

kduret:
Ok then let's set timeout to 10s

Indeed, we should test components independently, especially dialogs.
We'll handle global testing with cypress.

PR #8694, opened 2020-05-06
enh(refacto): Use context to split up resource components

bdauria:
## Description

This PR is a first step to an attempt to optimize Resource unit tests by splitting them up by dedicated Component.

To do that, a global Resource context has been introduced, through the use of the `useResourceContext()` hook. 

This will allow every major Component (Listing, Filter, Details, Actions) to be rendered independently, using a mocked version of the context. 

The follow up step will be to split up the index.test.tsx file into the corresponding Components. 

## Type of change

## Target serie

- [x] 20.04.x (master)

Make sure the whole Resource page is working as before (Listing, actions, dialogs, details, graphes).


package.json
```diff
@@ -95,7 +95,7 @@
     "install": "^0.13.0",
     "jsdom": "^16.2.2",
     "loaders.css": "^0.1.2",
-    "moment": "2.24.0",
+    "moment": "2.23.0",
```
kduret, line 98: why do you downgrade moment ?

bdauria:
> Tested and everything works fine except :
> When I disable autorefresh, the refresh button becomes disabled (but still clickable)
> It should stay enabled

This is now fixed

PR #8697, opened 2020-05-07
fix(APIv1): fix authorize check on webservices added directly in container

kduret:
## Description

Check webservice athorization by running authorize method
Catch error in authorize method to allow to throw specific exception (eg: invalid license)

**Fixes** BAM-1005

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)

Check APIv1

### DIFF
--- a/www/api/class/webService.class.php
+++ b/www/api/class/webService.class.php
@@ -361,6 +361,15 @@ class CentreonWebService
         }
     }
 
+    /**
+     * Check webservice authorization
+     *
+     * @param WebserviceAutorizePublicInterface|WebserviceAutorizeRestApiInterface $webservice
+     * @param string $action The action name
+     * @param CentreonUser|null $user The current user
+     * @param boolean $isInternal If the api is call from internal
+     * @return void
+     */
     private static function checkWebserviceAuthorization($webservice, $action, $user, $isInternal): void
     {
         $accessDenied = true;
--- a/www/api/class/webService.class.php
+++ b/www/api/class/webService.class.php
@@ -376,7 +376,7 @@ class CentreonWebService
     {
         $allowed = false;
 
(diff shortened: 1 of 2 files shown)


www/api/class/webService.class.php
```diff
@@ -347,36 +347,48 @@ public static function router(\Pimple\Container $dependencyInjector, $user, $isI
...
+            $wsObj::sendResult($data, 200, $resultFormat);
+        } catch (RestException $e) {
+            $wsObj::sendResult($e->getMessage(), $e->getCode());
+        } catch (Exception $e) {
+            $wsObj::sendResult($e->getMessage(), 500);
+        }
+    }
+
+    /**
+     * Check webservice authorization
```
callapa, line 365: Be careful, your method doesn't only check authorization, it also does redirections.
I suggest to change the name of your class like this: isWebserviceHasAllAuthorizations and to return TRUE or FALSE.
Then, depending on the status returned by this method you do the redirection.

www/api/class/webService.class.php
```diff
@@ -379,4 +362,27 @@ public static function router(\Pimple\Container $dependencyInjector, $user, $isI
...
+     * @param string $action The action name
+     * @param CentreonUser|null $user The current user
+     * @param boolean $isInternal If the api is call from internal
+     * @return boolean if the webservice is allowed for the current user
+     */
+    private static function isWebserviceAllowed($webservice, $action, $user, $isInternal): bool
+    {
+        $allowed = false;
+
+        // if implemented public interface skip checks
```
sc979, line 379: skip checks if public interface is implemented

PR #8708, opened 2020-05-11
fix(BE): remove popup errors in logs when altering http request

sc979:
## Description

FIX DT and ACK details popin

**Fixes** # (MON-5399)

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [x] 19.04.x
- [x] 19.10.x
- [ ] 20.04.x (master)

## Checklist

### DIFF
--- a/www/include/monitoring/downtime/xml/broker/makeXMLForDowntime.php
+++ b/www/include/monitoring/downtime/xml/broker/makeXMLForDowntime.php
@@ -120,23 +120,21 @@ if (false === $svcId) {
 }
 
 $rowClass = "list_one";
-if (isset($res)) {
-    while ($row = $res->fetch()) {
-        $row['comment_data'] = strip_tags($row['comment_data']);
-        $xml->startElement('dwt');
-        $xml->writeAttribute('class', $rowClass);
-        $xml->writeElement('author', $row['author']);
-        $xml->writeElement('start', $row['actual_start_time']);
-        if (!$row['fixed']) {
-            $row['end_time'] = (int)$row['actual_start_time'] + (int)$row['duration'];
-        }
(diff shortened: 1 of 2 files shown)


www/include/monitoring/downtime/xml/broker/makeXMLForDowntime.php
```diff
@@ -120,21 +120,23 @@
...
 $rowClass = "list_one";
-while ($row = $stmt->fetch()) {
-    $row['comment_data'] = strip_tags($row['comment_data']);
-    $xml->startElement('dwt');
-    $xml->writeAttribute('class', $rowClass);
-    $xml->writeElement('author', $row['author']);
-    $xml->writeElement('start', $row['actual_start_time']);
-    if (!$row['fixed']) {
-        $row['end_time'] = (int)$row['actual_start_time'] + (int)$row['duration'];
+if (isset($res)) {
```
kduret, line 123: why do you check $res ? It should always exist

jeremyjaouen: Same expectation than kduret , 
when initiate $res should be a PDOStatement instance if prepare() method went well, or false if something wrong happened. Then it should be true or false as result of execute() method. 
So in every case $res is set but you should test if $res === true

sc979: I forgot to remove this line from debug. This line isn't required anymore since I added the host_id check on line 73 in the last PR ;)
@jeremyjaouen -> before the modifications, if host_id was not set, the PDO failed when binding false on host_id in the query of the second case

PR #8716, opened 2020-05-12
feat(ui): Hide / show graph line in Event details view

bdauria:
## Description

This adds the possibility of hiding / showing lines on the performance graph (only on the details panel):

toggable-legend

## Type of change

## Target serie

- [x] 20.04.x (master)


Moujimouja:
yep, no doubt that a checkbox is not what users want. Name should be clickable (as 100% graph libs do)

bdauria:
@kduret @Moujimouja the new behaviour with latest changes is as follow:

hideshow

kduret:
Since there is no more checkbox, I think the colored circle can be retrieved to the left of the metric label

Moujimouja:
> Since there is no more checkbox, I think the colored circle can be retrieved to the left of the metric label

agreed, do not change the circle position

bdauria:
@Moujimouja @kduret Legend icon moved back to the left hand side:

Screenshot from 2020-05-13 19-58-05

Thebarda:
LGTM with latest changes 👍 

graph1

graph2

PR #8728, opened 2020-05-18
feat(ui): Use Context API to provide userContext and add ACL

bdauria:
## Description
This provides a base for the userContext via the Context API. This will, afterwards, replace the current store to clean up the architecture.

For now, the userContext is populated with the following content:

Screenshot from 2020-05-18 15-45-57

## Type of change

## Target serie
- [x] 20.04.x (master)


www/front_src/src/Provider/index.tsx
```diff
@@ -0,0 +1,81 @@
...
+        store.dispatch(setLocale(retrievedUser.locale.slice(0, 2)));
+
+        setDataLoaded(true);
+      })
+      .catch((error) => {
+        if (pathEq(['response', 'status'], 401)(error)) {
+          window.location.href = 'index.php?disconnect=1';
+        }
+      });
+  }, []);
```
Thebarda, line 59: There is a "lot" of requests made here. I think it is better to display a loading progress to telle to the user that something is processing, especially for low connection

sc979:
I've rebased the branch and added a spin to the LanguageSelection.feature acceptance test

www/front_src/src/Provider/index.test.tsx
```diff
@@ -0,0 +1,86 @@
...
+
+const retrievedTranslations = {
+  en: {
+    hello: 'Hello',
+  },
+};
+
+let userContext: UserContext | null = null;
+
+jest.mock('../App', () => {
```
kduret, line 39: it could be interesting to extract this mock in a separate file, cause I think this will be usefull for other unit tests

Thebarda:
LGTM with latest changes. 

contextProvider

PR #8736, opened 2020-05-22
feat(ui): Use user context in resources to handle ACL

bdauria:
## Description

This uses the previously setup userContext to query ACL in order to prevent the user from executing actions he's not allowed to:

Screenshot from 2020-05-22 10-06-13
Screenshot from 2020-05-22 10-06-20

Bulk actions with mixed ACL are still allowed, but a warning message is displayed to let the user know that not all resources will be affected:

Screenshot from 2020-05-22 10-06-48
Screenshot from 2020-05-22 10-06-56

## Type of change

## Target serie
- [x] 20.04.x (master)

- Manipulate ACL to restrict actions to either hosts or services
- Make sure not allowed actions are not allowed (corresponding commands disabled)
- Make sure actions with mixed ACL are allowed, with a warning message.


lang/es_ES.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -14494,3 +14494,9 @@ msgstr ""
 
 msgid "See Documentation"
 msgstr ""
+
+msgid "Services will not be affected because you don't have sufficient permission"
+msgstr ""
+
+msgid "Hosts will not be affected because you don't have sufficient permission"
+msgstr ""
```
sc979, line 14502:
```suggestion
msgid "Services will not be affected because you don't have sufficient permission"
msgstr ""

#~ msgid "Hosts will not be affected because you don't have sufficient permission"
#~ msgstr ""
```

sc979: Comment missing translations to avoid to display empty strings instead of english

lang/pt_BR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -14941,3 +14941,9 @@ msgstr ""
 
 msgid "See Documentation"
 msgstr ""
+
+msgid "Services will not be affected because you don't have sufficient permission"
+msgstr ""
+
+msgid "Hosts will not be affected because you don't have sufficient permission"
+msgstr ""
```
sc979, line 14949: please, comment empty strings

lang/pt_PT.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -14955,3 +14955,9 @@ msgstr ""
 
 msgid "See Documentation"
 msgstr ""
+
+msgid "Services will not be affected because you don't have sufficient permission"
+msgstr ""
+
+msgid "Hosts will not be affected because you don't have sufficient permission"
+msgstr ""
```
sc979, line 14963: please, comment empty strings

features/bootstrap/LanguageSelectionContext.php
```diff
@@ -13,7 +13,6 @@ class LanguageSelectionContext extends CentreonContext
     public function theUserWithAutologinEnabled()
     {
         $this->currentPage = new ParametersMyAccountPage($this);
-        $this->currentPage->save();
```
sc979, line 16: Why do you remove the form submit ?

Thebarda:
LGTM with latest changes
Capture d’écran (20)
Capture d’écran (21)
Capture d’écran (19)


I have one remark about error message when clicking on disabled 'check' button.

Capture d’écran (22)

I think it would be better for user to add a more explicit message

features/bootstrap/LanguageSelectionContext.php
bdauria: This step is supposed to be about the visit of the page, the saving step doesn't make any sense here because the following ones are doing assertions.
This messed up the tests because the assertions were done in a too short of a time frame.

sc979: indeed, it's the given step. but why is it randomly breaking the TA since 10 days ?

bdauria:
@Thebarda That's definitely not a normal behaviour, nothing should happen when the button is clicked. 

I suggest that I address the issue in another PR so that we don't further block the changes.

PR #8748, opened 2020-05-27
WIP chore(npm): Update packages

bdauria:
## Description
This updates packages to latest versions and fixes resulting typing errors.

## Type of change

## Target serie

- [x] 20.04.x (master)


www/front_src/src/Provider/index.tsx
```diff
@@ -39,7 +39,7 @@ const AppProvider = (): JSX.Element | null => {
     Promise.all([
       getUser(userEndpoint),
       getTranslations(translationEndpoint),
-      getAcl(aclEndpoint),
+      getAcl('http://localhost:5000/mock/acl'),
```
jdelpierre, line 42: Isn't it better to capture this information in a configuration or constant to be reused?

PR #8749, opened 2020-05-27
fix(ui): Prevent disabled actions resulting from limited ACL to be clickable

bdauria:
## Description
This prevents the User with limited ACL on an action to execute it even if the button is disabled. 

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 20.04.x (master)

- Set a limited ACL for the current User on either acknowledge, downtime or check

- Go to the event view page

- try to click on a resource inline action (single Resource) corresponding to the limited ACL, that should be disabled (either acknowledge, downtime or check) 

Expected result: No action is triggered.

### DIFF
--- a/www/front_src/src/Resources/ActionButton.tsx
+++ b/www/front_src/src/Resources/ActionButton.tsx
@@ -21,7 +21,6 @@ type Props = {
 
 const ActionButton = ({
   title,
-  onClick,
   ariaLabel,
   disabled,
   ...props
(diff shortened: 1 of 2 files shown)


Thebarda:
LGTM. Icon buttons are correctly disabled
Capture d’écran (24)

www/front_src/src/Resources/ActionButton.tsx
```diff
@@ -34,7 +33,6 @@ const ActionButton = ({
         <IconButton
           className={classes.button}
           color="primary"
-          onClick={onClick}
           disabled={disabled}
```
Thebarda, line 36: Maybe we could remove `disabled` because it is forwarded by `...prop`, isn't it ?

PR #8750, opened 2020-05-28
enh(ui): Use default font in Performance graph

bdauria:
## Description

This PR makes the performance graph use Centreon default font (Roboto). It affects the X and Y axes, as well as the Tooltip legend (title and legend are formatted using mui's Typography, so they are already properly formatted).

localhost_4000_centreon_monitoring_events

## Type of change

## Target serie

- [x] 20.04.x (master)

- Go to the event views
- Open a performance graph details
- Check that Roboto is used consistently (compare with the demo platform if in doubt)

### DIFF
--- a/www/front_src/src/Resources/Graph/Performance/Lines.tsx
+++ b/www/front_src/src/Resources/Graph/Performance/Lines.tsx
@@ -5,6 +5,8 @@ import { pipe, uniq, prop, map } from 'ramda';
 
 import { fade } from '@material-ui/core';
 
+import { fontFamily } from '.';
+
 const getGraphLines = ({ lines, formatValue }): Array<JSX.Element> => {
   const getUnits = (): Array<string> => {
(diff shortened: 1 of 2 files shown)


Thebarda:
LGTM, Roboto font is correctly used on performance graph tab and graph tooltip

Capture d’écran (26)
Capture d’écran (25)

One remark: There are scrollbars on tooltip graph. This happens on Chrome (without zoom)

bdauria:
> LGTM, Roboto font is correctly used on performance graph tab and graph tooltip
> 
> Capture d’écran (26)
> Capture d’écran (25)
> 
> One remark: There are scrollbars on tooltip graph. This happens on Chrome (without zoom)

The scroll are due to the fixed size of the tooltip and the dynamic nature of the legend height. We will probably see how we handle the rendering better in the near future

bdauria:
Changes accidentally included in: https://github.com/centreon/centreon/pull/8751

PR #8761, opened 2020-06-01
fix(ui): Fix lower value display of Event view performance graph 

bdauria:
## Description

This fixes the lower limit check so that even if the lower limit is set to 0, graph with only 0 values are displayed anyway:

localhost_4000_centreon_monitoring_events (6)

## Type of change

## Target serie
- [x] 20.04.x (master)

- Open a performance graph (such as Centreon-ping)
- wait that enough values are available
- Check that a line (for instance Packet-loss for Centreon-ping) is displayed


Thebarda:
LGTM with latest changes
Capture d’écran (31)

PR #8762, opened 2020-06-02
enh(api): update iconNormalizer to make it generic

kduret:
## Description

Update iconNormalizer to make it generic.
It will allows to use it in other endpoints which need to expose icon

**Fixes** BAM-1024

## Type of change

## Target serie

- [ ] 2.8.x
- [ ] 18.10.x
- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x (master)


src/Centreon/Application/Normalizer/IconUrlNormalizer.php
```diff
@@ -34,43 +33,23 @@ class IconUrlNormalizer implements ContextAwareNormalizerInterface
...
-        // normalize resource icon
-        if ($resource->getIcon() !== null) {
-            $this->normalizeIcon($resource->getIcon());
-        }
-
-        // normalize parent resource icon
-        if ($resource->getParent() !== null && $resource->getParent()->getIcon() !== null) {
-            $this->normalizeIcon($resource->getParent()->getIcon());
+        if (isset($_SERVER['REQUEST_URI']) && preg_match('/^(.+)\/api\/.+/', $_SERVER['REQUEST_URI'], $matches)) {
+            $icon->setUrl($matches[1] . self::IMG_DIR . '/' . $icon->getUrl());
```
sc979, line 42: shouldn't you check that matches[1] is found ? eg: REQUEST_URI does not contains the word "api"

kduret: It cannot enter in this condition if $matches[1] does not exist ;)
Cause preg_match returns false if nothing is matched

PR #8780, opened 2020-06-08
fix(ui): disable service checkbox when not allowed

kduret:
## Description

Disable "applies to services checkbox when action is not allowed on services : 
- downtime dialog
- acknowledgement dialog

I've also reduced size of checkboxes (small)

image

image

**Fixes** MON-5150

## Type of change

* log in with a user which cannot acknowledge and set downtime on services
* go to events view and check a host in listing
* try to acknowledge or set a downtime using top action buttons
* check that checkbox which apply action on services is not checked and disabled

## Target serie

- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

### DIFF
--- a/www/front_src/src/Resources/Actions/Resource/Acknowledge/Dialog.tsx
+++ b/www/front_src/src/Resources/Actions/Resource/Acknowledge/Dialog.tsx
@@ -107,7 +107,7 @@ const DialogAcknowledge = ({
               control={
                 <Checkbox
                   checked={values.acknowledgeAttachedResources}
-                  disabled={!canAcknowledgeServices}
+                  disabled={!canAcknowledgeServices()}
                   inputProps={{ 'aria-label': labelAcknowledgeServices }}
                   color="primary"
                   onChange={handleChange('acknowledgeAttachedResources')}
--- a/www/front_src/src/Resources/Actions/Resource/Acknowledge/Dialog.tsx
+++ b/www/front_src/src/Resources/Actions/Resource/Acknowledge/Dialog.tsx
@@ -106,7 +106,10 @@ const DialogAcknowledge = ({
             <FormControlLabel
               control={
                 <Checkbox
-                  checked={values.acknowledgeAttachedResources}
+                  checked={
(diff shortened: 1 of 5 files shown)


www/front_src/src/Resources/Actions/index.test.tsx
```diff
@@ -557,4 +558,48 @@ describe(Actions, () => {
...
+      cannotDowntimeServicesAcl,
+    ],
+    [
+      labelAcknowledge,
+      labelAcknowledge,
+      labelAcknowledgeServices,
+      cannotAcknowledgeServicesAcl,
+    ],
+  ])(
+    'disables services propagation option when trying to %p on hosts and not being able on services',
```
bdauria, line 576:
```suggestion
    'disables services propagation option when trying to %p on hosts when ACL on services are not sufficient,
```

www/front_src/src/Resources/Actions/Resource/Acknowledge/Dialog.tsx
```diff
@@ -78,42 +86,39 @@ const DialogAcknowledge = ({
...
+        <Grid item>
+          <FormControlLabel
+            control={
               <Checkbox
+                checked={values.notify}
                 inputProps={{ 'aria-label': labelNotify }}
                 color="primary"
-                value={values.notify}
                 onChange={handleChange('notify')}
+                size="small"
```
bdauria, line 97: I think we have enough space within the Acknowledge and Downtime dialog to display the checkboxes in their original sizes (it adds some comfort as well). 

```suggestion
```

kduret: are we sure we want to keep normal size ? cause we already use TextField as "small" in those dialogs

jdelpierre:
User with ability to ACK hosts and services: **All is OK**
image


**User with ability to ACK hosts but not services**
I checked a service and a host and try ACK or DT and you get a gray checkbox.

image
----
image

PR #8826, opened 2020-07-02
fix(secu): upgrade kind-of dependency

sc979:
## Description

Upgrade kind-of dependency to latest version (6.0.3)

@kduret , @bdauria Caution, there's still one package reported as highly vulnerable :
image

**Fixes** # (dependabot)

## Type of change

## Target serie

- [x] 19.04.x
- [ ] 19.10.x
- [ ] 20.04.x
- [ ] 20.10.x (master)

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,6 +51,7 @@
     "connected-react-router": "^6.5.2",
     "dom-serializer": "^0.1.1",
     "install": "^0.13.0",
+    "kind-of": "^6.0.3",
     "loaders.css": "^0.1.2",
     "moment-timezone": "^0.5.26",
     "numeral": "^2.0.6",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -51,6 +51,7 @@
     "connected-react-router": "^6.5.2",
     "dom-serializer": "^0.1.1",
     "install": "^0.13.0",
+    "kind-of": "^6.0.3",
```
bdauria, line 54: I don't think we use kind-off anywhere in the codebase. Did you use the command `npm run audit fix` to attempt to fix them?

sc979: audit fix, did not fix nothing. When you open the package-lock.json, you'll find that kind-of is required as a dependency for multiple dependencies we use, like : is-data-descriptor, is-descriptor, is-number, merge-deep, micromatch, ... 
So we don't use it directly, but the modules we need, do

sc979: @bdauria , as mentioned in the description, there's still (on the 19.10.x and 19.04.x branches) One high vulnerability reported by npm audit : 
image
This PR isn't intended to fix it. I let you two decide what to do and how to resolve it :

PR #8860, opened 2020-07-17
feat(ui): Edit Events view custom filters

bdauria:
## Description

This adds the possibility to edit custom filters via the RightPanel:

edit-filters

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 20.04.x
- [x] 20.10.x (master)


www/front_src/src/Resources/Filter/Edit/index.tsx
```diff
@@ -0,0 +1,76 @@
+import React from 'react';
```
kduret, line 1:
```suggestion
import * as React from 'react';
```

www/front_src/src/Resources/Filter/Edit/index.tsx
```diff
@@ -0,0 +1,76 @@
...
+    <div className={classes.header}>
+      <Typography variant="h5" align="center">
+        {labelEditFilters}
+      </Typography>
+    </div>
+  );
+
+  if (!editPanelOpen) {
+    return null;
+  }
```
kduret, line 64:
```suggestion

  if (!editPanelOpen) {
    return null;
  }

  const closeEditPanel = (): void => {
    setEditPanelOpen(false);
  };

  const Sections = [
    {
      expandable: false,
      id: 'edit',
      Section: (
        <div className={classes.filters}>
          {customFilters?.map((filter) => (
            <EditFilterCard key={filter.id} filter={filter} />
          ))}
        </div>
      ),
    },
  ];

  const Header = (
    <div className={classes.header}>
      <Typography variant="h5" align="center">
        {labelEditFilters}
      </Typography>
    </div>
  );
```

kduret: Diff is big, but I just moved `!editPanelOpen` condition on top to avoid unnecessary treatment

lang/fr_FR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -15424,4 +15424,22 @@ msgid "Filter created"
...
+msgstr "Supprimer le filtre?"
+
+msgid "Filter deleted"
+msgstr "Filtre supprimé"
+
+msgid "Filter updated"
+msgstr "Filtre actualisé"
+
+msgid "Name cannot be empty"
+msgstr "Le nom ne peut être vide"
```
kduret, line 15445:
```suggestion
msgstr "Le nom ne peut pas être vide"
```

bdauria:
> According to the story :
> **When I get out of the field OR type enter Then the name is saved**
> 
> So you should remove the save iconButton in the edit panel

Done

lang/fr_FR.UTF-8/LC_MESSAGES/messages.po
```diff
@@ -15424,4 +15424,22 @@ msgid "Filter created"
 msgstr "Filtre créé"
 
 msgid "Filter saved"
-msgstr "Filtre sauvegardé"
\ No newline at end of file
+msgstr "Filtre sauvegardé"
```
kduret, line 15427: Do you think `enregistré` is a better translation than `sauvegardé` ?

bdauria: I wouldn't say one is better than the other, but we should be consistent in its use.

www/front_src/src/Resources/Filter/Edit/EditFilterCard.tsx
```diff
@@ -0,0 +1,198 @@
...
+
+      loadCustomFilters();
+    });
+  };
+
+  const cancelDelete = (): void => {
+    setDeleting(false);
+  };
+
+  const loading = isNil(customFilters);
```
Thebarda, line 131:
```suggestion
  const customFiltersEmpty = isNil(customFilters);
```
This is not a loading state but this indicates that `customFilters` is null or undefined

www/front_src/src/Resources/index.tsx
```diff
@@ -59,9 +60,7 @@ const Resources = (): JSX.Element => {
 
   const { selectedDetailsEndpoints } = detailsContext;
 
-  if (isNil(filterContext.customFilters)) {
-    return <CircularProgress className={classes.loadingIndicator} />;
-  }
+  const loading = isNil(filterContext.customFilters);
```
Thebarda, line 63:
```suggestion
  const customFiltersEmpty = isNil(filterContext.customFilters);
```
Same comment here

jdelpierre:
I tested it, but it's not intuitive in the filters panel the saving "auto" at behavior of the field => exit it or key press enter on keyboard.

I prefer the behavior on the video it must be clear with a "save button" at end of the line.

And another remark, if you delete the current filter, the filters selector on the listing page become "white" :( Maybe I think it's return at the "default custom value" => "New Filter"

kduret:
> I tested it, but it's not intuitive in the filters panel the saving "auto" at behavior of the field => exit it or key press enter on keyboard.
> 
> I prefer the behavior on the video it must be clear with a "save button" at end of the line.
> 
> And another remark, if you delete the current filter, the filters selector on the listing page become "white" :( Maybe I think it's return at the "default custom value" => "New Filter"

The idea was to do somthing like jira ticket title.
A label field (not textfield) which show mouse `text` cursor on hover
Once it's clicked, the label becomes a textfield.
Once click away or enter is pressed, the textfield becomes back to a label with saved value

bdauria:
@kduret @jdelpierre I will implement it in that way

jdelpierre:
> > I tested it, but it's not intuitive in the filters panel the saving "auto" at behavior of the field => exit it or key press enter on keyboard.
> > I prefer the behavior on the video it must be clear with a "save button" at end of the line.
> > And another remark, if you delete the current filter, the filters selector on the listing page become "white" :( Maybe I think it's return at the "default custom value" => "New Filter"
> 
> The idea was to do somthing like jira ticket title.
> A label field (not textfield) which show mouse `text` cursor on hover
> Once it's clicked, the label becomes a textfield.
> Once click away or enter is pressed, the textfield becomes back to a label with saved value

That's exactly what we rethought with Bruno :) 
lined up guys

bdauria:
@Thebarda @kduret @jdelpierre new behaviour:

new-behavior

Thebarda:
> @Thebarda @kduret @jdelpierre new behaviour:
> 
> new-behavior
> 

I really like this one

PR #8877, opened 2020-07-27
fix(dashboard): human readable date for reporting timeline detail popin

adr-mo:
## Description

This PR aims to display correctly the day displayed in the reporting timeline detail popin.

**/!\ IMPORTANT /!\**

Translation has not been handled since it is not handled for the whole timeline...

**Before**

3ad82c40-6b99-46fd-82eb-35070377a5ed

**After**

image

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 19.04.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

- Go to `Reporting  >  Dashboard  >  Host Groups` for example (servicegroups and hosts should be checked too)
- Click on a part of the timeline
- The detailed popin should be displayed and the "Day: date" should be human readable

## Checklist

- [ ] I followed the **coding style guidelines** provided by Centreon

### DIFF
--- a/www/include/reporting/dashboard/xmlInformations/common-Func.php
+++ b/www/include/reporting/dashboard/xmlInformations/common-Func.php
@@ -92,7 +92,7 @@ function fillBuffer($statesTab, $row, $color)
     $detailPopup = '{table class=bulleDashtab}';
     $detailPopup .= '{tr}{td class=bulleDashleft colspan=3}' . $Day .
         ': {span class ="isTimestamp isDate"}';
-    $detailPopup .= date('Y-m-d', $date_start) . '{/span} --  ' . $Duration . ': ' .
+    $detailPopup .= $date_start . '{/span} --  ' . $Duration . ': ' .
         CentreonDuration::toString($totalTime) . '{/td}{td class=bulleDashleft }' . $Alert . '{/td}{/tr}';
     foreach ($statesTab as $key => $value) {
         $detailPopup .= '{tr}' .
--- a/www/include/reporting/dashboard/xmlInformations/common-Func.php
+++ b/www/include/reporting/dashboard/xmlInformations/common-Func.php
@@ -92,7 +92,7 @@ function fillBuffer($statesTab, $row, $color)
     $detailPopup = '{table class=bulleDashtab}';
     $detailPopup .= '{tr}{td class=bulleDashleft colspan=3}' . $Day .
         ': {span class ="isTimestamp isDate"}';
-    $detailPopup .= date('Y-m-d', $date_start) . '{/span} --  ' . $Duration . ': '.
+    $detailPopup .= date('Y-m-d', $date_start) . '{/span} --  ' . $Duration . ': ' .
         CentreonDuration::toString($totalTime) . '{/td}{td class=bulleDashleft }' . $Alert . '{/td}{/tr}';
     foreach ($statesTab as $key => $value) {
         $detailPopup .= '{tr}' .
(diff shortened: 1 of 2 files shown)


www/include/reporting/dashboard/xmlInformations/common-Func.php
```diff
@@ -92,7 +92,7 @@ function fillBuffer($statesTab, $row, $color)
     $detailPopup = '{table class=bulleDashtab}';
     $detailPopup .= '{tr}{td class=bulleDashleft colspan=3}' . $Day .
         ': {span class ="isTimestamp isDate"}';
-    $detailPopup .= $date_start . '{/span} --  ' . $Duration . ': '.
+    $detailPopup .= date('Y-m-d', $date_start) . '{/span} --  ' . $Duration . ': '.
```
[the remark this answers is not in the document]
sc979: These labels classes : "isTimestamp" & "isDate" were used to manage the localization and was working since the 18.10.
IMHO, you should check why moment is unable to convert the timestamp

www/include/reporting/dashboard/xmlInformations/common-Func.php
```diff
@@ -92,7 +92,7 @@ function fillBuffer($statesTab, $row, $color)
     $detailPopup = '{table class=bulleDashtab}';
     $detailPopup .= '{tr}{td class=bulleDashleft colspan=3}' . $Day .
         ': {span class ="isTimestamp isDate"}';
-    $detailPopup .= $date_start . '{/span} --  ' . $Duration . ': '.
+    $detailPopup .= date('Y-m-d', $date_start) . '{/span} --  ' . $Duration . ': ' .
```
sc979, line 95: "timestamp" and "isDate" are classes keyword used by moment to overide and localize the timestamp in the customer format.
IMHO, you should try to find why moment isn't able to convert it, as it was introduced and working on the 18.10

adr-mo: Are you sure about this @sc979 ?
I've been testing it in the latest 18.10.x, 19.04.x, 19.10.x and 20.04.x and it does not work :cry:

sc979: Check the ticket MON-2491

adr-mo: Sometimes validation can fail :+1: 
The detail popin was a subtask we might have skip its validation. With the actual code it cannot work. 

**Why ?**

Currently the formatDateMoment() function searchs in the DOM for elements defined with the `isTimestamp` class

formatDateMoment() --
```
----
    //creating a collection of occurrences
    jQuery(".isTimestamp").each(function (index, element) {
        var myElement = jQuery(element);
---
```
When loading the timeline **ie** when selecting an `id` in the SELECT (a hostgroup id for instance) the popins are not part of the DOM. So the `formatDateMoment()` will not format the popin.

We have checked with @kduret

We've tried to add a setTimeout after the loadXML, to wait for data to be loaded to call the `formatDateMoment()`

This is **not OK**
- The date gets translated after the time defined for the timeout which is weird
- The date gets translated only one time (when opening the popup). If we try to reopen the popin the date do not get translated.

To sum up at the Timeline level it is currently not possible to add eventListeners so we can't do this on the JS / FRONT side.

It needs to be done on backend side

adr-mo: On the backend side, we will have to get the locale and Timezone on the PHP level (information available in the localStorage which can't be used on PHP).

We will have to get those values from the PHP session I guess.

WDYT @sc979 @kduret does it worth to do it knowing that none of the values are localized in the Timeline ... ?

sc979: Except that, this feature was working fine. I've tested it before QA.
As you say, under certain conditions, you're able to make it work.
IMHO : you should investigate to find what introduced this regression

kduret: A part of timeline library was overrided by Centreon, so finally I've updated this section to manage date conversion from frontend

sc979: You rock

PR #8891, opened 2020-08-04
fix(stats): get stats even if a module fails

kduret:
## Description

get statistics of a module even if it fails

**Fixes** MON-5624

## Type of change

## Target serie

- [ ] 19.04.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

Throw an error in a module statistics class, and check if it works

### DIFF
--- a/www/class/centreonStatsModules.class.php
+++ b/www/class/centreonStatsModules.class.php
@@ -1,6 +1,7 @@
 <?php
+
 /*
- * Copyright 2005-2019 Centreon
+ * Copyright 2005-2020 Centreon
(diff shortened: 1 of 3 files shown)


www/class/centreonStatsModules.class.php
```diff
@@ -1,6 +1,7 @@
 <?php
+
 /*
- * Copyright 2005-2019 Centreon
+ * Copyright 2005-2020 Centreon
  * Centreon is developped by : Julien Mathis and Romain Le Merlus under
```
sc979, line 5:
```suggestion
 * Centreon is developed by : Julien Mathis and Romain Le Merlus under
```

Issue #8868, opened 2020-07-22
PHP Warning: strftime(): Invalid date.timezone value 'CEST', we selected the timezone 'UTC' for now. in /usr/share/centreon/vendor/smarty/smarty/libs/Smarty_Compiler.class.php on line 395


Phoenixoboros:
# BUG REPORT INFORMATION

### Prerequisites

> The opened issue, must be code related. GitHub is not meant for support. Feel free to check the CONTRIBUTING section for more details.

***Versions***

For the RPM based systems

-- Copy/Paste the result of the following command --
```
centreon-broker-cbmod-20.04.6-3.el7.centos.x86_64
centreon-gorgone-20.04.3-1.el7.centos.noarch
centreon-plugin-Applications-Monitoring-Centreon-Map4-Jmx-20200602-094050.el7.centos.noarch
centreon-widget-servicegroup-monitoring-20.04.0-5.el7.centos.noarch
centreon-widget-host-monitoring-20.04.3-2.el7.centos.noarch
centreon-engine-daemon-20.04.4-1.el7.centos.x86_64
centreon-engine-20.04.4-1.el7.centos.x86_64
centreon-plugin-Network-Cisco-Standard-Snmp-20200602-094050.el7.centos.noarch
centreon-plugin-Applications-Protocol-Ldap-20200602-094050.el7.centos.noarch
centreon-web-20.04.4-6.el7.centos.noarch
centreon-widget-tactical-overview-20.04.0-5.el7.centos.noarch
centreon-trap-20.04.4-6.el7.centos.noarch
centreon-release-20.04-1.el7.centos.noarch
centreon-connector-20.04.0-2.el7.centos.x86_64
centreon-clib-20.04.0-7.el7.centos.x86_64
centreon-plugin-Applications-Protocol-Dns-20200602-094050.el7.centos.noarch
centreon-plugin-Hardware-Printers-Generic-Snmp-20200602-094050.el7.centos.noarch
centreon-plugin-Applications-Protocol-Ftp-20200602-094050.el7.centos.noarch
centreon-plugin-Applications-Databases-Mysql-20200602-094050.el7.centos.noarch
centreon-broker-storage-20.04.6-3.el7.centos.x86_64
centreon-license-manager-common-20.04.2-1.el7.centos.noarch
centreon-plugin-Operatingsystems-Linux-Snmp-20200602-094050.el7.centos.noarch
centreon-plugin-Applications-Protocol-Http-20200602-094050.el7.centos.noarch
[...]

PR #8905, opened 2020-08-11
fix(install): fix timezone getter in unattended script

kduret:
## Description

Fix unattended.sh when timezone name is not compatible with php

**Fixes** MON-5376

## Type of change

## Target serie

- [ ] 19.04.x
- [ ] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

Run unattended.sh with specific timezones


Nohzoh:
wouldn't it be interesting to accept a TZ environment variable in order to force the timezone ?
then we could imagine that the download web site provide explicitly the user timezone in the command line provided to execute the unattended script

kduret:
> wouldn't it be interesting to accept a TZ environment variable in order to force the timezone ?
> then we could imagine that the download web site provide explicitly the user timezone in the command line provided to execute the unattended script

Yes we could, but I think it's more interesting to do this automatically, cause I guess most f the users want to synchronize the centreon timezone with the server timezone.
But indeed, we could add the possibility to add an argument for that

jeremyjaouen (approved):

unattended.sh
```diff
@@ -114,10 +114,19 @@ print_step_end
...
+        $timezoneName = $matches[1];
+    }
+
+    if (date_default_timezone_set($timezoneName) === false) {
+      $timezoneName = "UTC";
+    }
+
+    echo $timezoneName;
+' 2> /dev/null`
 echo "date.timezone = $timezone" > /etc/opt/rh/rh-php72/php.d/10-centreon.ini
```
sc979, line 130: Caution : this file may not exists : 
https://docs.centreon.com/current/en/upgrade/upgrade-from-19-04.html#update-the-php-version

Nohzoh: I think that this script is done to make an install from scratch.
It should not be executed on a machine that already has a Centreon installed.

If this requirement is not met, many things could fail while executing this script.

sc979 (approved):

cgagnaire:
Closes https://github.com/centreon/centreon/issues/8868

PR #8948, opened 2020-08-28
fix(secu): correct ACL on returned user list

sc979:
## Description

Returns only allowed user list on the internal API, called from the share view feature instead of the whole user list on the platform

**Fixes** # (MON-5530)

## Type of change

- [x] Patch fixing an issue (non-breaking change)
- [ ] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software
- [ ] Updating documentation (missing information, typo...)

## Target serie

- [x] 19.04.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

<h2> How this pull request can be tested ? </h2>

Please contact me

PR #8951, opened 2020-08-31
fix(secu): correct ACL on returned group list

sc979:
## Description

Returns only allowed groups list on the internal API, called from the share view feature instead of the whole user list on the platform

**Fixes** # (MON-5530)

## Type of change

## Target serie

- [x] 19.04.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x (master)

Please contact me

### DIFF
--- a/www/class/centreonCustomView.class.php
+++ b/www/class/centreonCustomView.class.php
@@ -1081,7 +1081,7 @@ class CentreonCustomView
             /**
              * Find users linked to the view
              */
-            $userList = array();
+            $userList = [];
             if (empty($allowedContactIds)) {
                 return $userList;
             }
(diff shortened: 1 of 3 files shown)


sc979:
PR rebased on https://github.com/centreon/centreon/pull/8948

www/class/centreonCustomView.class.php
```diff
@@ -1109,30 +1109,62 @@ public function getUsersFromViewId(int $viewId): array
...
-            $stmt = $this->db->prepare($query);
-            $stmt->bindParam(':viewId', $viewId, PDO::PARAM_INT);
+            $allowedGroupIds = '';
+            foreach (array_keys($aclListOfGroupIds) as $groupId) {
+                // result's concatenation
+                if ('' !== $allowedGroupIds) {
+                    $allowedGroupIds .= ', ';
+                }
+                if (false !== filter_var($groupId, FILTER_VALIDATE_INT)) {
+                    $allowedGroupIds .= $groupId;
```
callapa, line 1142: Please join this two conditions to avoid the following case: ',,'

callapa (changes requested):

www/class/centreonCustomView.class.php
callapa: It would be best to use an array instead of a string.

callapa (approved):

loiclau (approved):

PR #8967, opened 2020-09-08
fix(resources-status): Fix Timeline infinite scroll showing incoherent results

bdauria:
## Description
This fixes some weird behavior while loading Timeline events and using infinite scroll.

## Type of change

## Target serie

- [x] 20.04.x
- [x] 20.10.x (master)

- Select a Resource with more than 10 events
- Select the corresponding Timeline tab in the details panel
- Do some guerilla loading / testing
- Check that the behavior is coherent and the result is always properly loaded

### DIFF
--- a/www/front_src/src/Resources/Details/tabs/Timeline/index.tsx
+++ b/www/front_src/src/Resources/Details/tabs/Timeline/index.tsx
@@ -90,28 +90,26 @@ const TimelineTab = ({ links }: Props): JSX.Element => {
     decoder: listTimelineEventsDecoder,
   });
 
-  const listTimeline = (
-    { toPage } = { toPage: page },
-  ): Promise<TimelineListing> => {
-    const getSearch = (): SearchParameter | undefined => {
-      if (isEmpty(selectedTypes)) {
-        return undefined;
-      }
+  const getSearch = (): SearchParameter | undefined => {
+    if (isEmpty(selectedTypes)) {
+      return undefined;
+    }
 
-      return {
-        lists: [
-          {
-            field: 'type',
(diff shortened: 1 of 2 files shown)


Thebarda (approved):

www/front_src/src/Resources/Details/tabs/Timeline/index.tsx
```diff
@@ -95,42 +111,49 @@ const TimelineTab = ({ links }: Props): JSX.Element => {
...
       parameters: {
-        page,
+        page: toPage,
         limit,
         search: getSearch(),
       },
     }).then((retrievedListing) => {
       const { meta } = retrievedListing;
       setTotal(meta.total);
+      setLoadingMoreEvents(false);
```
kduret, line 121: I think it's better to put it in a `finally` to reset it even if there is a request error

kduret (approved):

Thebarda (dismissed):

jdelpierre:
PR tested, so a new behaviour was suggested during the test. So waiting for new developments to test again.

kduret (commented):
Is it ready for review ?
Or you may need to set it as draft

bdauria:
> Is it ready for review ?
> Or you may need to set it as draft

it's ready for review now

kduret (approved):

jdelpierre (approved):

jdelpierre:
Test ok.
When you have more than 30 items, the scrolling loads the next ones and stops the timeline refresh.
When I come back to the top, a button appears to restart the "refresh" and go back to the "auto" state with the last items.

image

PR #9018, opened 2020-09-21
enh(resource-status): Make details tab URL query parameters interpretation more robust

bdauria:
## Description

This prevents the page to blow up if URL query parameters are invalid. 

To test: just set an invalid details parameter in the URL (either wrong resource id / parent id / type / parent type or invalid tab). The page shouldn't blow up. In case of wrong tab: the details tab is displayed instead. In case of wrong resource parameter: the panel is closed and an error message is displayed.  

**Fixes** # (issue)

## Type of change

## Target serie
- [x] 20.10.x (master)


Thebarda (approved):

jdelpierre:
It's OK for me, i have write an ID negative and this alert it's displaying

image

jdelpierre (approved):

PR #9064, opened 2020-10-01
fix(ui): fix ack & downtime tooltips on hover

kduret:
## Description

fix ack & downtime tooltips on hover

**Fixes** MON-6080

## Type of change

## Target serie

- [ ] 19.04.x
- [ ] 19.10.x
- [ ] 20.04.x
- [x] 20.10.x (master)

check downtime and acknowledgement tooltips

### DIFF
--- a/www/front_src/src/Resources/models.ts
+++ b/www/front_src/src/Resources/models.ts
@@ -32,9 +32,8 @@ export interface Resource {
   icon?: Icon;
   parent?: Parent;
   status: Status;
-  downtime_endpoint?: string;
+  links: ResourceLinks;
   acknowledged: boolean;
-  acknowledgement_endpoint?: string;
   in_downtime: boolean;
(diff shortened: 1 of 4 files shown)


www/front_src/src/Resources/models.ts
```diff
@@ -71,15 +64,18 @@ export interface Acknowledgement {
 }
 
 export interface ResourceEndpoints {
-  details: string;
-  performanceGraph?: string;
-  timeline: string;
+  details: string | null;
```
bdauria, line 67: Why use null instead of undefined here?

kduret: because this parameter is **always** returned by the API and is null or string

bdauria:
LGTM & tested

Thebarda (approved):

PR #9095, opened 2020-10-08
fix(resource-status): Avoid double details refresh and prevents auto refresh from loading the wrong details

bdauria:
## Description

This avoids resources being loaded twice and fixes an issue while changing the selected details which loads the previous one. 

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 20.04.x
- [x] 20.10.x (master)

### DIFF
--- a/www/front_src/src/Resources/Listing/useLoadResources/index.ts
+++ b/www/front_src/src/Resources/Listing/useLoadResources/index.ts
@@ -29,6 +29,7 @@ const useLoadResources = (): LoadResources => {
     customFilters,
     loadDetails,
     details,
+    selectedResourceId,
   } = useResourceContext();
 
   const refreshIntervalRef = React.useRef<number>();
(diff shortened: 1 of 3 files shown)


www/front_src/src/Resources/Listing/useLoadResources/index.ts
```diff
@@ -65,14 +67,20 @@ const useLoadResources = (): LoadResources => {
...
     loadDetails();
   };
 
   const initAutorefresh = (): void => {
     window.clearInterval(refreshIntervalRef.current);
 
     const interval = enabledAutorefresh
-      ? window.setInterval(load, refreshIntervalMs)
+      ? window.setInterval(() => {
+          load();
```
Thebarda, line 82: The old implementation has the same behavior, right ?

bdauria: yes

kduret (approved):

jdelpierre (approved):

PR #9099, opened 2020-10-08
fix(API): enhance hostname and name check

sc979:
## Description

Allow spaces on host
More accurate check for RFC compliant check

**Fixes** # (MON-6093)

## Type of change

## Target serie

- [ ] 19.04.x
- [ ] 19.10.x
- [ ] 20.04.x
- [x] 20.10.x (master)

### DIFF
--- a/www/include/Administration/parameters/DB-Func.php
+++ b/www/include/Administration/parameters/DB-Func.php
@@ -1051,7 +1051,7 @@ function updateRemoteAccessCredentials($db, $form, $centreonEncryption): void
     unset($ret['apiCredentials']);
 
     //convert values
-    $ret['apiSelfSignedCertificate'] = $ret['apiSelfSignedCertificate'] == 1 ? 'yes' : 'no';
+    $ret['apiPeerValidation'] = (int) $ret['apiPeerValidation'] === 1 ? 'no' : 'yes';
 
     //update information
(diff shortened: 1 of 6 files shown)


www/include/Administration/parameters/DB-Func.php
```diff
@@ -1051,7 +1051,7 @@ function updateRemoteAccessCredentials($db, $form, $centreonEncryption): void
     unset($ret['apiCredentials']);
 
     //convert values
-    $ret['apiSelfSignedCertificate'] = $ret['apiSelfSignedCertificate'] == 1 ? 'yes' : 'no';
+    $ret['apiPeerValidation'] = (int) $ret['apiPeerValidation'] === 1 ? 'no' : 'yes';
```
kduret, line 1054: `1` means `no`?

sc979: it's odd indeed. When checked the value is '1' for -> 
image

But when checked, "allow self signed certificate" means do not check "peer validation"
so yes, 1 means no in DB :)

kduret: ok :)
so the name of the form input is not good

kduret: let's update this later

loiclau (approved):

PR #9190, opened 2020-11-09
MON-6228 unattended centos8 support

Nohzoh:
## Description

Enhance unattended.sh script for Centos 8.
The scripts now checks which version of Centos is running and does what has to be done depending on this version.

Also avoid version number redondances in the script in order to simplify later updates.

## Type of change

## Target serie

- [ ] 19.10.x
- [ ] 20.04.x
- [x] 20.10.x
- [ ] 21.04.x (master)

Make a fresh install on a Centos 8 machine and check everything is installed as expected.
As a non regression test: do the same from a Centos 7 machine, check everything is ok.

### DIFF
--- a/unattended.sh
+++ b/unattended.sh
@@ -38,6 +38,7 @@ case "$rhrelease" in
     PHP_BIN="/opt/rh/rh-php72/root/bin/php"
     PHP_ETC="/etc/opt/rh/rh-php72/php.d"
     OS_SPEC_SERVICES="rh-php72-php-fpm httpd24-httpd"
+    PKG_MGR="yum"
     ;;
   '8')
     # CentOS 8 specific part
@@ -48,6 +49,7 @@ case "$rhrelease" in
     PHP_BIN="/bin/php"
     PHP_ETC="/etc/php.d"
     OS_SPEC_SERVICES="php-fpm httpd"
+    PKG_MGR="dnf"
     ;;
   *)
     error_and_exit "This unattended installation script only supports CentOS 7 and CentOS 8. Please check https://documentation.centreon.com/$CENTREON_MAJOR_VERSION/en/installation/introduction.html for alternative installation methods."
@@ -98,12 +100,12 @@ fi
 #
 
 print_step_begin "Centreon official repositories installation"
-yum -q clean all
+$PKG_MGR -q clean all
 
 if [[ $rhrelease == 7 ]]; then
   rpm -q centos-release-scl > /dev/null 2>&1
   if [ "x$?" '!=' x0 ] ; then
-    yum -q install -y centos-release-scl
+    $PKG_MGR -q install -y centos-release-scl
     if [ "x$?" '!=' x0 ] ; then
       error_and_exit "Could not install Software Collections repository (package centos-release-scl)"
     fi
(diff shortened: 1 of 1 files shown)


unattended.sh
```diff
@@ -25,15 +27,29 @@ print_step_begin "System analysis"
...
 if [ \! -e /etc/redhat-release ] ; then
-  error_and_exit "This unattended installation script only supports Red Hat compatible distributions. Please check https://documentation.centreon.com/20.10/en/installation/introduction.html for alternative installation methods."
+  error_and_exit "This unattended installation script only supports Red Hat compatible distributions. Please check https://documentation.centreon.com/$CENTREON_MAJOR_VERSION/en/installation/introduction.html for alternative installation methods."
 fi
-rhrelease=`cat /etc/redhat-release`
+rhrelease=$(rpm -E %{rhel})
 case "$rhrelease" in
-  'CentOS Linux release 7.'*)
+  '7')
+    # Good to go.
```
sc979, line 35: this comment does not clarify the intention
Good to go -> Begin CentOs7 script

unattended.sh
```diff
@@ -25,15 +27,29 @@ print_step_begin "System analysis"
...
+    RELEASE_RPM_URL="http://yum.centreon.com/standard/$CENTREON_MAJOR_VERSION/el7/stable/noarch/RPMS/centreon-release-$CENTREON_MAJOR_VERSION-2.el7.centos.noarch.rpm"
```
sc979, line 35: I'm not sure about the hardcoded '-2'. A variable may help maintain the link

unattended.sh
```diff
@@ -25,15 +27,29 @@ print_step_begin "System analysis"
...
+    PHP_BIN="/opt/rh/rh-php72/root/bin/php"
+    PHP_ETC="/etc/opt/rh/rh-php72/php.d"
+    OS_SPEC_SERVICES="rh-php72-php-fpm httpd24-httpd"
+    ;;
+  '8')
     # Good to go.
```
sc979: same

unattended.sh
Nohzoh: I remove the comment
It was in the previous code to indicate the only case when the rest of the script could be executed
Now there are 2 cases, so indeed il doesn't clarify the intention anymore

unattended.sh
```diff
@@ -25,15 +27,29 @@ print_step_begin "System analysis"
...
+    dnf -y install dnf-plugins-core epel-release
+    dnf -y update gnutls
+    dnf config-manager --set-enabled PowerTools
+    RELEASE_RPM_URL="http://yum.centreon.com/standard/$CENTREON_MAJOR_VERSION/el8/stable/noarch/RPMS/centreon-release-$CENTREON_MAJOR_VERSION-2.el8.noarch.rpm"
```
sc979, line 44: same

unattended.sh
```diff
@@ -25,15 +27,29 @@ print_step_begin "System analysis"
...
+    PHP_BIN="/bin/php"
+    PHP_ETC="/etc/php.d"
+    OS_SPEC_SERVICES="php-fpm httpd"
     ;;
   *)
-    error_and_exit "This unattended installation script only supports CentOS 7. Please check https://documentation.centreon.com/20.10/en/installation/introduction.html for alternative installation methods."
+    error_and_exit "This unattended installation script only supports CentOS 7. Please check https://documentation.centreon.com/$CENTREON_MAJOR_VERSION/en/installation/introduction.html for alternative installation methods."
```
sc979, line 50: centOs7 and centOs8

unattended.sh
Nohzoh: It was already hardcoded like that before.
The centreon-release rpm is usually not update early after the release so it should almost never change.
Do you have an idea to avoir this ?

unattended.sh
Nohzoh: same also

unattended.sh
Nohzoh: same also ;)

unattended.sh
Nohzoh: true
fixed

adr-mo (approved):

sc979 (approved):

PR #9194, opened 2020-11-10
feat(resource-status/Listing): Add an URL column

bdauria:
## Description

Displays an URL column redirecting to the given resource URL.

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 20.10.x
- [x] 21.04.x (master)

- Define a 'notes' URL on a host or service
- Check that the URL column is displayed (it shouldn't be for resources without any notes URL)
- Check that the button redirects to the given URL


Thebarda:
notes_url
It seems Link icon is not displayed while my resource actually has a 'notes' URL.
By given a look at your code, you are calling to `links.externals.notes` instead of `links.externals.notes_url`

Thebarda (commented):
notes_url
It seems Link icon is not displayed while my resource actually has a 'notes' URL.
By given a look at your code, you are calling to `links.externals.notes` instead of `links.externals.notes_url`

www/front_src/src/Resources/translatedLabels.ts
```diff
@@ -176,3 +176,4 @@ export const labelSubmit = 'Submit';
 export const labelStatusSubmitted = 'Status submitted';
 export const labelOutput = 'Output';
 export const labelNoResourceFound = 'No resource found';
+export const labelUrl = 'URL';
```
Thebarda, line 179: Do not forget the translation

bdauria: There is no translation for URL

Thebarda:
I have several issues by using this feature :
- Link icon is displayed even if I haven't defined note URL. The API returns a notes_url equal to "" instead of `null`or `undefined`
https://www.loom.com/share/759a48a575bd4443aff4d2b43c587a07
In this video, I have defined a note URL only for my host.
- When I click on the icon, it opens the resource's details then it redirects to the correct URL
https://www.loom.com/share/4ae5ffbe07be4f77a03f9e7a573faea5

bdauria:
> I have several issues by using this feature :
> 
> * Link icon is displayed even if I haven't defined note URL. The API returns a notes_url equal to "" instead of `null`or `undefined`
>   https://www.loom.com/share/759a48a575bd4443aff4d2b43c587a07
>   In this video, I have defined a note URL only for my host.
> * When I click on the icon, it opens the resource's details then it redirects to the correct URL
>   https://www.loom.com/share/4ae5ffbe07be4f77a03f9e7a573faea5

fixed

www/front_src/src/Resources/Listing/columns/Url.tsx
```diff
@@ -0,0 +1,40 @@
...
+      onClick={(e): void => {
+        e.stopPropagation();
+      }}
+    >
+      <IconButton
+        title={labelUrl}
+        ariaLabel={labelUrl}
+        onClick={(): null => {
+          return null;
+        }}
```
kduret, line 32: could not be something like `() => {}` instead of returning null ?

kduret (approved):

www/front_src/src/Resources/Listing/columns/Url.tsx
Thebarda: Empty arrow function are not allowed by ESLint

www/front_src/src/Resources/Listing/columns/index.tsx
```diff
@@ -239,6 +240,15 @@ export const getColumns = ({ actions, t }): Array<Column> => [
...
     width: 200,
   },
+  {
+    id: 'url',
+    label: '',
+    type: ColumnType.component,
+    getRenderComponentOnRowUpdateCondition: T,
+    Component: UrlColumn,
+    sortable: false,
+    width: 50,
```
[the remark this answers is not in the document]
bdauria: Unfortunately this doesn't work, since clickable: true will prevent from redirecting to the endpoint

Thebarda (approved):
Tested & LGTM

PR #9204, opened 2020-11-17
feat(resource-status/Details): Add FQDN and Alias tiles in the details panel

bdauria:
## Description

This adds two new tiles to the details panel: Alias and FQDN / Address (displayed when the corresponding values are defined):

Screenshot from 2020-11-17 09-36-03

## Type of change

## Target serie

- [x] 20.10.x
- [x] 21.04.x (master)

- Define a FQDN / Address to a resource
- Define an Alias
- Export the configuration
- Make sure the corresponding tiles are displayed
- Make sure the tiles are not displayed when the values are not defined


kduret (approved):

jdelpierre:
In the domain IT, the term "address" related to FQDN is "address ip" or not?

jdelpierre (approved):

bdauria:
> In the domain IT, the term "address" related to FQDN is "address ip" or not?

It can be an address or a FQDN (the domain name) so we have to display both terms.

jdelpierre:
> > In the domain IT, the term "address" related to FQDN is "address ip" or not?
> 
> It can be an address or a FQDN (the domain name) so we have to display both terms.

Ok I understand, but what does "address" look like? An IP? 
Because for me it's a postal address :/

Thebarda (commented):
FQDN
Maybe it can be done i another ticket. It depend on how bid the deal is : it may be better to display very long FQDN/Alias, which are without breakable spaces, with a truncate suffix. 
Otherwise, everything looks good to me

bdauria:
> FQDN
> Maybe it can be done i another ticket. It depends on how big the deal to handle is : it may be better to display very long FQDN/Alias, which are without breakable spaces, with a truncate suffix.
> Otherwise, everything looks good to me

@Thebarda I've reduced the fontSize, hopefully it'll cover the majority of the cases

kduret (approved):

jdelpierre (approved):

Thebarda (approved):
Tested & LGTM

PR #9257, opened 2020-11-30
fix(secu): enhance default http configuration when using sources

sc979:
## Description

disable http trace option and prevent from click-jacking in http configuration set by default when user use the sources to install his platform

**Fixes** # (MON-4751 & MON-4750)

## Type of change

## Target serie

- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x
- [x] 21.04.x (master)

### DIFF
--- a/libinstall/functions
+++ b/libinstall/functions
@@ -1394,6 +1394,11 @@ prepare_apache_config() {
 
     # Prepare apache config
     ${CAT} << __EOT__ > $directory/centreon.apache.conf
+Header set X-Frame-Options: "sameorigin"
+ServerSignature Off
+ServerTokens Prod
+TraceEnable Off
+
 Alias /centreon/api $INSTALL_DIR_CENTREON
 Alias /centreon $INSTALL_DIR_CENTREON/www/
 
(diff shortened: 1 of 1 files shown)


libinstall/functions
```diff
@@ -1394,6 +1394,11 @@ prepare_apache_config() {
 
     # Prepare apache config
     ${CAT} << __EOT__ > $directory/centreon.apache.conf
+Header set X-Frame-Options: "sameorigin"
+ServerSignature Off
+ServerTokens Prod
+TraceEnable Off
```
Nohzoh, line 1400: The apache documentation indicates that enabling TRACE method does not expose any security vulnerability.
https://httpd.apache.org/docs/2.4/en/mod/core.html#TraceEnable
Why should it be explicitly disabled?

callapa: IMHO, the TraceEnable option should not be used in production mode, but only for debugging purposes.

jeremyjaouen (approved):

kduret (approved):

PR #9296, opened 2020-12-14
feat(Resource-status/Graph): Add tooltip to add a comment in a Resource performance graph

bdauria:
## Description

This adds a way to add a comment by clicking anywhere on the performance graph:

Screenshot from 2020-12-14 17-14-32
Screenshot from 2020-12-14 17-14-40

## Type of change

## Target serie
- [x] 21.04.x (master)


www/front_src/src/Resources/Listing/columns/Graph/index.tsx
```diff
@@ -57,7 +57,9 @@ const GraphColumn = ({
           <PerformanceGraph
             endpoint={endpoint}
             graphHeight={150}
+            resource={row}
             timeline={[]}
+            onAddComment={(): void => undefined}
```
Thebarda, line 62: Why did you add a `onAddComment` property that implement any logic ?

Thebarda (changes requested):

Thebarda:
Capture d’écran (78)
 I have two remarks about the user experience : 
First, it could be great to display a pointer cursor over the graph to tell the user that he can click on the graph / do something.
Secondly, when displaying the popover box, add a shadow on that box to improve readability between graph and the popover box

www/front_src/src/Resources/Listing/columns/Graph/index.tsx
bdauria: It's needed in case the timeline is displayed, so that the comment is added in real time instead of waiting for it to be retrieved by the API

bdauria:
@Thebarda There is now a little bit of shadow around the comment button, and the cursor is of type 'crosshair'

Thebarda (commented):
There is an empty file `www/front_src/src/Resources/Graph/Performance/Graph/Annotations/Area/index.tsx`

kduret (dismissed):
IMHO, it should be usefull to display the time information beside the "add comment" button to know at which time exactly we'll add the comment

jdelpierre (changes requested):
It's ok, but so remove this file before approve ;) 
www/front_src/src/Resources/Graph/Performance/Graph/Annotations/Area/index.tsx

bdauria:
> IMHO, it should be usefull to display the time information beside the "add comment" button to know at which time exactly we'll add the comment

I thought about it as well, but the tooltip above metrics displays it already, and I'm not sure about the usefulness of adding a comment on non existing metric values (where the tooltip doesn't display).

jdelpierre (dismissed):

kduret (dismissed):

Thebarda (dismissed):

bdauria:
@kduret I added the time anyway: 

Screenshot from 2020-12-15 11-46-13

Thebarda (approved):

jdelpierre (approved):

PR #9318, opened 2020-12-16
fix(test): remove phpunit warnings

kduret:
## Description

* remove phpunit warnings : 
```
There were 2 warnings:

1) Tests\Centreon\Application\Controller\PlatformTopologyControllerTest::testDeletePlatformTopologySuccess
Trying to configure method "deletePlatform" which cannot be configured because it does not exist, has not been specified, is final, or is static

2) Tests\Centreon\Domain\PlatformTopology\PlatformTopologyServiceTest::testDeletePlatformTopologySuccess
Method deletePlatform may not return value of type NULL, its return declaration is ": void"
```
* take into account failed builds for quality gate : to be able to reset it

## Type of change

## Target serie

- [ ] 19.10.x
- [ ] 20.04.x
- [ ] 20.10.x
- [x] 21.04.x (master)

### DIFF
--- a/tests/php/Centreon/Application/Controller/PlatformTopologyControllerTest.php
+++ b/tests/php/Centreon/Application/Controller/PlatformTopologyControllerTest.php
@@ -290,10 +290,6 @@ class PlatformTopologyControllerTest extends TestCase
 
     public function testDeletePlatformTopologySuccess(): void
     {
-        $this->platformTopologyService
-            ->expects($this->once())
-            ->method('deletePlatform')
-            ->willReturn(null);
         $platformTopologyController = new PlatformTopologyController($this->platformTopologyService);
         $platformTopologyController->setContainer($this->container);
 
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -86,6 +86,7 @@ try {
         recordIssues(
           enabledForFailure: true,
           aggregatingResults: true,
+          ignoreFailedBuilds: false,
           qualityGates: [[threshold: 1, type: 'NEW', unstable: false]],
           tools: [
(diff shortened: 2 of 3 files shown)


tests/php/Centreon/Application/Controller/PlatformTopologyControllerTest.php
```diff
@@ -290,10 +290,6 @@ public function testGetPlatformJsonGraph(): void
 
     public function testDeletePlatformTopologySuccess(): void
     {
-        $this->platformTopologyService
```
sc979, line 293: what kind of warning is returned when this test is runned ?

sc979: wouldn't removing this part make the test useless ?
as the test is meant to test the deletePlatform method

kduret: this method returns void, not null
So we don't need to override the default mock

kduret:
```
There were 2 warnings:

1) Tests\Centreon\Application\Controller\PlatformTopologyControllerTest::testDeletePlatformTopologySuccess
Trying to configure method "deletePlatform" which cannot be configured because it does not exist, has not been specified, is final, or is static

2) Tests\Centreon\Domain\PlatformTopology\PlatformTopologyServiceTest::testDeletePlatformTopologySuccess
Method deletePlatform may not return value of type NULL, its return declaration is ": void"
```

jeremyjaouen (approved):

adr-mo (approved):

loiclau (approved):

PR #9325, opened 2020-12-17
feat(Resource-status/Services): Display graphs in the Services details tab

bdauria:
## Description
This allows the user to switch to a 'Graph mode' within the Services details tab in order to display the performance graphs. 

It also saves some parameters in the URL such as the mode (graph / list) and the selected time period. 

Note that this mode will evolve in the near future and that this is only a first step. 

graphMode

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 21.04.x (master)


Thebarda:
Tested with the latest changes and it seems there is a bug that add the first graphs twice in the list

https://www.loom.com/share/83474e70a41a4ba8b6d7f2a4b2f77f78

www/front_src/src/Resources/Details/tabs/Services/index.tsx
```diff
@@ -71,81 +55,133 @@ const ServicesTab = ({ details }: TabProps): JSX.Element => {
...
+            />
+          ) : (
+            <ServiceList
+              services={entities}
+              onSelectService={selectService}
+              infiniteScrollTriggerRef={infiniteScrollTriggerRef}
+            />
+          );
+        }}
+      </InfiniteScroll>
```
Thebarda, line 181: Here, the infinitescroll's page is not resetted when switching the mode.

Thebarda (commented):
The code looks good and the feature works correctly with some service associated to a host. I'm looking for the conflict resolution to fully approve this PR.
Note : I've duplicated the "Ping" service 30 times and in graph mode, the 'scrolling' lag increase as far as I scroll. This is due to the amount of elements displayed. To be discussed with the front-end team.

EDIT : Here is a video that explains why it lags when scrolling through services graph : https://www.loom.com/share/bdb989e139f54432b772e905f7fc93c5

bdauria:
> The code looks good and the feature works correctly with some service associated to a host. I'm looking for the conflict resolution to fully approve this PR.
> Note : I've duplicated the "Ping" service 30 times and in graph mode, the 'scrolling' lag increase as far as I scroll. This is due to the amount of elements displayed. To be discussed with the front-end team.
> 
> EDIT : Here is a video that explains why it lags when scrolling through services graph : https://www.loom.com/share/bdb989e139f54432b772e905f7fc93c5

Services graphs are now memoized through their resource ids, tooltipX and interval in order to increase their resilience against infinite scrolling. 

One issue now remains: the infinite scroll might be inconsistent with the number of service graphs because we filter the services on the presence of graph data. This will be addressed in another PR

Thebarda (dismissed):

www/front_src/src/Resources/Graph/Performance/Graph/index.tsx
```diff
@@ -79,6 +79,8 @@ interface Props {
   base: number;
   lines: Array<LineModel>;
   xAxisTickFormat: string;
+  tooltipX?: number;
+  onTooltipDisplay?: (tooltipX?: number) => void;
```
jdelpierre, line 83: it's the same variable ? or just same name?

jdelpierre (dismissed):

Thebarda (approved):

jdelpierre (approved):

PR #9340, opened 2020-12-23
enh(install): Add Oracle Linux support

quanghungb:
## Description
- Add some logging 
- Add Oracle Linux (7 & 8) support

## Type of change

- [X ] New functionality (non-breaking change)

## Target serie

- [ ] 19.10.x
- [ ] 20.04.x
- [X ] 20.10.x
- [X ] 21.04.x (master)

### DIFF
--- a/unattended.sh
+++ b/unattended.sh
@@ -1,178 +1,615 @@
-#!/bin/sh
+#!/bin/bash
 
-function error_and_exit {
-  echo -e "$1"
+### Define all supported constants
+OPTIONS=":t:v:r:l:"
+declare -A SUPPORTED_LOG_LEVEL=([DEBUG]=0 [INFO]=1 [WARNING]=2 [ERROR]=3)
+declare -A SUPPORTED_TOPOLOGY=([central]=1 [poller]=1)
+declare -A SUPPORTED_VERSION=([20.10]=1)
+declare -A SUPPORTED_REPOSITORY=([testing]=1 [unstable]=1 [stable]=1)
+default_timeout_in_sec=5
+script_short_name="$(basename $0)"
+###
+
+#Define default values
+topology=central 	#Default topology to be installed
+version=20.10		#Default version to be installed
+repo=stable			#Default repository to used
+operation=install	#Default operation to be executed
+runtime_log_level="INFO" #set default log level to INFO
+
+##FIXME - to be set dynmically & and support other tickets
+CENTREON_MAJOR_VERSION="20.10"
+CENTREON_RELEASE_VERSION="$CENTREON_MAJOR_VERSION-2"
+
+#Variables dynamically set
+os_release=
+os_version=
+
+# Variables will be defined later according to the target system OS
+BASE_PACKAGES=
+RELEASE_RPM_URL=
+PHP_BIN=
+PHP_ETC=
+OS_SPEC_SERVICES=
+PKG_MGR=
+has_systemd=
+CENTREON_REPO=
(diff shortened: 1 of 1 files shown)


lgcosta:
Just note to help to download ISO OracleLinux:

https://mirrors.dotsrc.org/oracle-linux/

* because using Oracle software delivery is very annoying !!!

unattended.sh
```diff
@@ -1,178 +1,615 @@
-#!/bin/sh
+#!/bin/bash
```
Nohzoh, line 1: The is somehow limitating.
Bash is not always available, particularly on minimal installs.

But I agree that the script is much simpler for bash than it would be for sh.
So we choose simplicity of the script over the compatibility with more Linux installs?

unattended.sh
```diff
@@ -1,178 +1,615 @@
...
+					;;
+					
+				: )
+					log "ERROR" "Option -"$OPTARG" requires an argument."
+					usage
+					exit 1
+					;;
+			esac
+	done
+    shift $((OPTIND-1))
```
Nohzoh, line 172: tabs everywhere instead of spaces
```suggestion
	shift $((OPTIND-1))
```

unattended.sh
```diff
@@ -1,178 +1,615 @@
...
+default_timeout_in_sec=5
+script_short_name="$(basename $0)"
+###
+
+#Define default values
+topology=central 	#Default topology to be installed
+version=20.10		#Default version to be installed
+repo=stable			#Default repository to used
+operation=install	#Default operation to be executed
+runtime_log_level="INFO" #set default log level to INFO
```
Nohzoh, line 18: aligned comments
```suggestion
#Define default values
topology=central 		#Default topology to be installed
version=20.10			#Default version to be installed
repo=stable			#Default repository to used
operation=install		#Default operation to be executed
runtime_log_level="INFO"	#set default log level to INFO
```

unattended.sh
```diff
@@ -1,178 +1,615 @@
...
+      fi
+		  ;;
 
+    8*) 
+      log "INFO" "Setting specific part for v8 ($os_version)"
+      
+      case "$os_release" in
+        centos-release* | centos-linux-release*)
+          BASE_PACKAGES=(dnf-plugins-core epel-release)
+          dnf config-manager --set-enabled powertools
```
Nohzoh, line 313: there is an issue here with CentOS version before 8.3, the repository id `PowerTools`

lgcosta: > there is an issue here with CentOS version before 8.3, the repository is `PowerTools` with upper cases

wouldn't it be interesting to apply (or recommend with a note) a `dnf update` before running the script routines ?

unattended.sh
```diff
@@ -1,178 +1,615 @@
...
+	  if [ "x$?" '=' x0 ] ; then
+		has_systemd=1
+	  else
+		has_systemd=0
+	  fi
+	else
+	  has_systemd=0
+	fi
+}
+#========= end of function is_systemd_present()
```
tuntoja, line 360: Could we be using this to check the OS is using systemd (or not) ?
`ps --no-headers -o comm1`

Nohzoh (approved):

Nohzoh:
The master branch is building the 21.04 version
So as is, the script won't work for master builds
This change should be at least backported to the 20.10 branch and also tested and perhaps backported to 20.04 and 19.10

quanghungb:
replaced by https://github.com/centreon/centreon/pull/9619

PR #9407, opened 2021-01-18
enh(Resource-status/Graph): optimize lisibility of the Graphs

bdauria:
## Description

This enhances the display and the user experience of the performance graphs for Resources:
- Displays a single metric by clicking on it (unselects all others)
- If the metric is the only one selected and is clicked, displays all metrics
- Toggles a metric by ctrl + clicking on it
- Enhances the tooltip metric display
- Allows the tooltip to be displayed outside the panel
- Optimize the length of the legend by truncating if it doesn't fit the panel width 

graph-enh

## Type of change

## Target serie

- [x] 21.04.x (master)


Thebarda (dismissed):

kduret (dismissed):

jdelpierre (dismissed):

jdelpierre:
I've tried with some graph data, and it's correct. 
image

www/front_src/src/Resources/Graph/Performance/Graph/MetricsTooltip.tsx
```diff
@@ -23,7 +24,7 @@ const useStyles = makeStyles((theme) => ({
     display: 'flex',
     flexDirection: 'column',
   },
-  dateTime: {
+  emphasized: {
```
jdelpierre, line 27: Why this term ? "emphasized" ?

kduret (approved):

www/front_src/src/Resources/Graph/Performance/Graph/MetricsTooltip.tsx
kduret: You can check emphasis here : https://material.io/design/color/text-legibility.html#text-backgrounds
in french, It's like "mettre en valeur"

jdelpierre: Yes ! ok thanks :)

jdelpierre (approved):

www/front_src/src/Resources/Graph/Performance/Graph/MetricsTooltip.tsx
bdauria: it means that the item is presented with more importance

PR #9415, opened 2021-01-21
fix(secu): use the cron to remove expired sessions

sc979:
## Description

When the browser is closed without login out, the session can remain valid.
The Cron will now check and clean the saved sessions

Resulting with a 3 min session validity value (set in centreon-ui form)
```
2020-11-13 11:22:02|1|0|0|[WEB] [192.168.0.16] Authentication succeeded for 'admin'
2020-11-13 11:25:01|0|0|0|[CRON] [127.0.0.1] Remove expired session 'admin'
```

**Fixes** # (MON-4253)

## Type of change

## Target serie

- [x] 2.8.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x
- [x] 21.04.x (master)

### DIFF
--- a/cron/session.php
+++ b/cron/session.php
@@ -1,0 +1,96 @@
+#!@PHP_BIN@
+<?php
+
+/*
+ * Copyright 2005-2021 Centreon
+ * Centreon is developed by : Julien Mathis and Romain Le Merlus under
+ * GPL Licence 2.0.
+ *
+ * This program is free software; you can redistribute it and/or modify it under
+ * the terms of the GNU General Public License as published by the Free Software
+ * Foundation ; either version 2 of the License.
+ *
+ * This program is distributed in the hope that it will be useful, but WITHOUT ANY
+ * WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A
+ * PARTICULAR PURPOSE. See the GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along with
+ * this program; if not, see <http://www.gnu.org/licenses>.
+ *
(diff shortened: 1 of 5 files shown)


callapa (dismissed):
This file is not intended to check sessions.
Please perform this check before reading the session in Centreon

cron/session.php
```diff
@@ -0,0 +1,79 @@
+#!@PHP_BIN@
```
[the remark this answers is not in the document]
sc979: unconsistent...

cron/session.php
```diff
@@ -0,0 +1,79 @@
...
+     * Remove expired sessions
+     */
+    // Find duration of session_expiration
+    $sessionDuration = SESSION_DEFAULT_DURATION;
+    $durationQuery = $pearDB->query("SELECT `value` from options where `key` = 'session_expire'");
+    if (($duration = $durationQuery->fetch(\PDO::FETCH_ASSOC)) !== false) {
+        $sessionDuration = $duration['value'];
+    } else {
+        // cannot find the default value. setting a new value
+        $pearDB->query("DELETE FROM options WHERE `key` = 'session_expire'");
```
kduret, line 45: why trying to delete a value if you have already checked it does not exist ?

kduret (commented):
did you find why the session expiration is not checked when opening centreon web ?

cron/session.php
sc979: I checked that the value doesn't exist.
But if the key does not too, I won't be able to update the value. And if the key exists I won't be able to insert a new value

I prefer to clean the data before setting a new default value.
But this case should not occurs unless the options table is corrupted

PR #9416, opened 2021-01-21
fix(Resource-status/Filter): Fix Filter overflow by wrapping the fields in case of resize

bdauria:
## Description

This optimizes the display a bit in order to fix the responsiveness of the page by automatically wrapping the fields in case of resize, since the overflow is now set to hidden. It also removes some unnecessary labels in order to clear the UI a little bit:

resize

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 21.04.x (master)


www/front_src/src/Resources/Filter/index.tsx
```diff
@@ -255,64 +246,70 @@ const Filter = (): JSX.Element => {
...
-            field="name"
-            fullWidth
-          />
-          <Button color="primary" onClick={clearAllFilters}>
-            {t(labelClearAll)}
-          </Button>
-        </div>
+        <ParentSize>
+          {({ width }): JSX.Element => {
+            const limitTags = width < 1000 ? 1 : 2;
```
kduret, line 251: We'll may need to improve this calculation, to add more than 2 tags when possible

Here, I've selected `Down`, `Up`, and `Ok`. All these tags should be displayed
image

kduret (approved):

www/front_src/src/Resources/Filter/index.tsx
bdauria: Yes, this can be improve. The thing we don't want though is having wrapped fields in case all filters are filled up

jdelpierre:
It's OK for me. 

Before
image


After
image

jdelpierre (approved):

jdelpierre:
ps :

` msgid "Clear all"`

This string is not used on backend ?

bdauria:
> ps :
> 
> ` msgid "Clear all"`
> 
> This string is not used on backend ?

Not that I know, i remmember having added it

Thebarda (approved):

PR #9421, opened 2021-01-24
enh(ui): increase timeline density

kduret:
## Description

* Increase timeline density
* update style to improve readability of essential information (content)
* Remove click listing on chips (+ avoid grey background on click)
* Display properly carriage return of event content

To be discussed with @bdauria 

before : 
image

after : 
image

Refs: MON-6650

## Type of change

## Target serie

- [ ] 19.10.x
- [ ] 20.04.x
- [ ] 20.10.x
- [x] 21.04.x (master)

Check timeline events in resource status view


bdauria:
This looks good :) 

Can we align the chip to the time ?

www/front_src/src/Resources/Details/tabs/Services/List.tsx
```diff
@@ -52,6 +52,7 @@ const ServiceList = ({
             <Paper className={classes.serviceCard}>
               <div className={classes.serviceDetails}>
                 <StatusChip
+                  clickable={false}
```
bdauria, line 55: I can’t really think of a place where a StatusChip would be clickable, so we wight want to set it to false by default in @centreon-ui

Thebarda:
This looks good ! What do you think to move the 'alias' chip to the right side ?

kduret:
> This looks good ! What do you think to move the 'alias' chip to the right side ?

I've tried but it's a bit weird on "events" cause there is already the `Tries` property on the right

www/front_src/src/Resources/Details/tabs/Services/List.tsx
kduret: I thought it was used by BAM for enabled/disabled icon, but it's not.
So I'll move default value to false

bdauria (approved):

Thebarda (approved):

PR #9431, opened 2021-01-28
fix(secu): remove expired sessions more globally

sc979:
## Description

When the browser is closed without login out or when a computer is back from hibernation, the session can remain valid.
The check on session expiration is made more globally when the session is checked
And more often while notifications are send and when internal API is called, to avoid to refresh an expired session

**Fixes** # (MON-4253)

## Type of change

## Target serie

- [x] 2.8.x
- [x] 19.10.x
- [x] 20.04.x
- [x] 20.10.x
- [x] 21.04.x (master)

### DIFF
--- a/www/class/centreonSession.class.php
+++ b/www/class/centreonSession.class.php
@@ -125,7 +125,6 @@ class CentreonSession
     /**
      * Delete all expired sessions
      * @param CentreonDB $db
-     * @throws Exception
      */
     public static function deleteExpiredSession(CentreonDB $db): void
     {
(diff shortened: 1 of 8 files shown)


www/class/centreonSession.class.php
```diff
@@ -118,6 +122,23 @@ public static function checkSession($sessionId, CentreonDB $db)
         return ($total > 0) ? 1 : 0;
     }
 
+    /**
+     * Delete all expired sessions
+     * @param CentreonDB $db
+     * @throws Exception
```
kduret, line 128:
```suggestion
```

kduret (approved):

jeremyjaouen (approved):

PR #9451, opened 2021-02-04
fix(Resource-status/Downtime): Fix navigator default language

bdauria:
## Description

This fixes the language displayed in the Downtime dialog. 

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 21.04.x (master)


Thebarda (dismissed):

jdelpierre (approved):

www/front_src/src/Resources/Actions/Resource/Downtime/Dialog.tsx
```diff
@@ -111,11 +118,11 @@ const DialogDowntime = ({
 
   class Adapter extends DayjsAdapter {
     public format(date, formatString): string {
-      return dayjs(date).tz(timezone).format(formatString);
+      return format({ date, formatString });
     }
 
     public date(value): dayjs.Dayjs {
```
kduret, line 124: Only `Dayjs` could be imported, instead of default import

bdauria: We still use the default import right below (see the next method), so we need to import it this way anyway.

kduret (approved):

PR #9514, opened 2021-02-24
enh(Resource-status/Listing): Improve Listing responsiveness

bdauria:
## Description

This takes advantage of the new Listing layout to make the columns more responsive, and prevent them from wrapping around:

## Type of change

## Target serie

- [x] 21.04.x (master)


kduret (commented):
remove `centreon-license-manager/.DS_Store` file

package.json
```diff
@@ -88,7 +88,7 @@
     "@material-ui/lab": "^4.0.0-alpha.57",
     "@material-ui/pickers": "^3.2.10",
     "@material-ui/styles": "^4.11.2",
-    "@visx/visx": "^1.3.0",
+    "@visx/visx": "1.4.0",
```
bdauria, line 91: Latest visx version looks broken, fix it for the moment

www/front_src/src/Resources/Listing/columns/index.tsx
```diff
@@ -210,6 +215,7 @@ const ParentResourceColumn = ({
 };
 
 interface ColumnsProps {
+  isSmallScreen?: boolean;
```
kduret, line 218: seems not used

Thebarda (dismissed):

kduret (dismissed):

Thebarda (dismissed):
Tested and LGTM

www/front_src/src/Resources/Listing/index.test.tsx
```diff
@@ -371,7 +373,9 @@ describe(Listing, () => {
 
     const chipLabel = `${acknowledgedEntity?.name} ${labelAcknowledged}`;
 
-    const chip = await findByLabelText(chipLabel);
+    const chip = await findByLabelText(chipLabel, undefined, {
+      timeout: 10000,
```
Thebarda, line 377: Jest tears down the test after 5 seconds, maybe the timeout for 10 seconds is not very relevant

bdauria: We changed the default timeout to 10sec to avoid ci failures

Thebarda (approved):

www/front_src/src/Resources/Listing/index.test.tsx
jdelpierre: why only the CI?  and why just x2 from the default?

jdelpierre (approved):

www/front_src/src/Resources/Listing/index.test.tsx
bdauria: It’s not just a change within the CI, it’s global, but the failures manifested themselves mostly on the CI

PR #9556, opened 2021-03-05
fix(Resource-status/Listing): Fix display of Resources with the same ID

bdauria:
## Type of change

## Target serie

- [x] 20.10.x
- [x] 21.04.x (master)


www/front_src/src/Resources/Listing/index.tsx
```diff
@@ -69,7 +70,17 @@ const ResourceListing = (): JSX.Element => {
 
   const resourceDetailsOpenCondition = {
     name: 'detailsOpen',
-    condition: ({ id }): boolean => equals(id, selectedResourceId),
+    condition: (row): boolean => {
+      const { id, parent } = row;
+      const parentId = parent?.id;
+
+      // console.log(selectedResourceId, selectedResourceParentId);
```
kduret, line 77: remove debug

www/front_src/src/Resources/Listing/index.tsx
```diff
@@ -103,6 +112,12 @@ const ResourceListing = (): JSX.Element => {
     SortOrder,
   ];
 
+  const getCompositeId = (row: Resource): [number, number] => {
+    const parentId = path<number>(['parent', 'id'], row);
+
+    return [parentId || row.id, row.id];
```
kduret, line 118: if the resource is a service with following ids : host_id => 1, service_id => 1
and another resource is a host with following ids : host_id => 1
composite ids will be the same : [1,1]

www/front_src/src/Resources/Listing/index.tsx
```diff
@@ -103,6 +112,12 @@ const ResourceListing = (): JSX.Element => {
     SortOrder,
   ];
 
+  const getCompositeId = (row: Resource): [number, number?] => {
```
kduret, line 115: I forgot it, but we'll integrate business activities.
So we can have collisions between host id and ba id cause both do not have parent.

IMHO, composite id cannot be an integer. It should be a generated hash from : type, id, parent_type, parent_id

Thebarda: A hash like UUID ? Should the backend generate this hash rather than the frontend ?

bdauria: I agree with @Thebarda, this heavy lifting work should be typically done by the backend. Actually this whole composite ID thing needs to be replaced by a "uuid" property coming from the resource and calculated using type, id and parent_id.

kduret: new property `uuid` will be returned by backend

adr-mo (dismissed):
I approve the backend part

adr-mo (dismissed):

src/Centreon/Domain/Monitoring/Resource.php
```diff
@@ -56,6 +55,11 @@ class Resource
     public const TYPE_SERVICE = 'service';
     public const TYPE_HOST = 'host';
 
+    /**
+     * @var string|null
```
callapa, line 59: A uuid cannot be null because it represents an existing system resource.
Therefore the uuid should be part of the constructor.

src/Centreon/Domain/Monitoring/ResourceService.php
```diff
@@ -289,6 +290,34 @@ public function replaceMacrosInExternalLinks(ResourceEntity $resource): void
         }
     }
 
+    /**
+     * Generate resource uuid from resource and its parent
+     *
+     * @param ResourceEntity $resource
+     * @return void
+     */
+    private function generateResourceUuid(ResourceEntity $resource): void
```
callapa, line 299: The Uuid must be defined by the resource itself.

callapa (changes requested):

kduret:
@callapa done

callapa (approved):

Thebarda (approved):

PR #9563, opened 2021-03-08
[20.04.x] fix(secu): Fix npm vulnerabilities

bdauria:
## Description

This updates vulnerable packages. 

## Type of change

## Target serie

- [x] 20.04.x


www/front_src/src/Resources/Graph/Performance/Lines.tsx
```diff
@@ -7,8 +7,17 @@ import { fade } from '@material-ui/core';
...
 
-const getGraphLines = ({ lines, base }): Array<JSX.Element> => {
+interface GraphLinesProps {
+  lines: Array<LineModel>;
+  base?: number;
+}
+
+const getGraphLines = ({
+  lines,
+  base,
```
Thebarda, line 19:
```suggestion
  base = 1000,
```
Same default prop value as `www/front_src/src/Resources/Graph/Performance/formatMetricValue/index.ts`

www/front_src/src/Resources/Listing/columns/State/DetailsTable/index.tsx
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable react/no-array-index-key */
```
Thebarda, line 1: Why this one ?

bdauria: Because there is an index used for a key

Thebarda (approved):

PR #9645, opened 2021-04-01
fix(Resource-status): Various fixes

bdauria:
## Description

This fixes:
- The columns configuration not being saved when leaving the page
- The possibility to unselect all columns (at least one must be selected)
- The Resources with only passive checks not being displayed in the "checks" column
- The visibility of Resources with enabled notification in the "netofiication" column
- Tre information column collapsing (minimum width set)

**Fixes** # (issue)

## Type of change

## Target serie

- [x] 21.04.x (master)

### DIFF
--- a/package.json
+++ b/package.json
@@ -83,7 +83,7 @@
     "webpack-merge": "^5.7.3"
   },
   "dependencies": {
-    "@centreon/ui": "github:centreon/centreon-ui#fix-listing-header-shift",
+    "@centreon/ui": "github:centreon/centreon-ui",
     "@centreon/ui-context": "github:centreon/ui-context",
     "@date-io/dayjs": "1.3.13",
     "@material-ui/core": "4.11.3",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "@babel/preset-env": "^7.13.12",
     "@babel/preset-react": "^7.12.13",
     "@babel/preset-typescript": "^7.13.0",
-    "@centreon/frontend-core": "github:centreon/frontend-core",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -24,7 +24,7 @@
     "@babel/preset-env": "^7.13.12",
     "@babel/preset-react": "^7.12.13",
     "@babel/preset-typescript": "^7.13.0",
-    "@centreon/frontend-core": "github:centreon/frontend-core",
+    "@centreon/frontend-core": "github:centreon/frontend-core#6079299b5e4b7608813bfa40df90f724ace6bbe9",
```
Thebarda, line 27: Will you take off this commit hash in another PR ?

Thebarda (approved):

www/front_src/src/Resources/Listing/columns/Checks.tsx
```diff
@@ -1,22 +1,45 @@
...
+  if (row.active_checks === false) {
     return (
-      <IconColumn>
-        <Tooltip title={labelChecksDisabled}>
-          <SyncDisabledIcon color="primary" fontSize="small" />
-        </Tooltip>
-      </IconColumn>
+      <Column Icon={SyncProblemIcon} title={t(labelOnlyPassiveChecksEnabled)} />
     );
   }
```
kduret, line 45: If passive checks and active checks are both enabled, no icon will be displayed ?
Is it what we expert ?

bdauria: Yes

kduret (approved):

package.json
bdauria: Yes, I’ll wait that all the current PRs are merged so that we don’t deal with gigantic conflicts

PR #9680, opened 2021-04-09
fix(notice) dont cast before empty check

loiclau:
## Description

dont cast before empty check

[07-Apr-2021 11:57:52 UTC] PHP Notice:  Undefined index: gorgone_api_username in /usr/share/centreon/src/Centreon/Domain/Gorgone/ConfigurationLoader.php on line 97
[07-Apr-2021 11:57:52 UTC] PHP Notice:  Undefined index: gorgone_api_password in /usr/share/centreon/src/Centreon/Domain/Gorgone/ConfigurationLoader.php on line 108

## Type of change

## Target serie

- [ ] 19.10.x
- [ ] 20.04.x
- [x] 20.10.x
- [x] 21.04.x (master)

### DIFF
--- a/src/Centreon/Domain/Gorgone/ConfigurationLoader.php
+++ b/src/Centreon/Domain/Gorgone/ConfigurationLoader.php
@@ -72,7 +72,7 @@ class ConfigurationLoader implements ConfigurationLoaderApiInterface
         if (!$this->isOptionsLoaded) {
             $this->loadConfiguration();
         }
-        return (string) $this->gorgoneParameters[self::GORGONE_API_ADDRESS] ?? null;
+        return $this->gorgoneParameters[self::GORGONE_API_ADDRESS] ?? null;
     }
 
     /**
@@ -83,7 +83,7 @@ class ConfigurationLoader implements ConfigurationLoaderApiInterface
         if (!$this->isOptionsLoaded) {
             $this->loadConfiguration();
         }
-        return (int) $this->gorgoneParameters[self::GORGONE_API_PORT] ?? null;
+        return $this->gorgoneParameters[self::GORGONE_API_PORT] ?? null;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Centreon/Domain/Gorgone/ConfigurationLoader.php
```diff
@@ -83,7 +83,7 @@ public function getApiPort(): ?int
         if (!$this->isOptionsLoaded) {
             $this->loadConfiguration();
         }
-        return (int) $this->gorgoneParameters[self::GORGONE_API_PORT] ?? null;
+        return $this->gorgoneParameters[self::GORGONE_API_PORT] ?? null;
```
kduret, line 86:
```suggestion
        return $this->gorgoneParameters[self::GORGONE_API_PORT]
            ? (int) $this->gorgoneParameters[self::GORGONE_API_PORT]
            : null;
```

loiclau: with php 7 return of funtion, :?int auto cast '1' to 1

kduret: centreon-bot agrees with me :)
furthermore, if we do that, don't cast `isApiConnectionSecure` to bool, cause we need consistency

kduret (approved):

src/Centreon/Domain/Gorgone/ConfigurationLoader.php
callapa: @loiclau 
Your remark is wrong because we use the strict mode of PHP.
Have you never noticed the **declare(strict_types=1)** command at the beginning of every PHP file?

sc979 (approved):

jeremyjaouen (approved):

PR #9743, opened 2021-04-27
chore(install): update to version 20.10.7

tuntoja:
Update version to 20.10.7 in files.


sc979 (approved):
insertBaseConf.sql is already bumped to 20.10.7

kduret (approved):

PR #9761, opened 2021-04-30
feat(unattended): add debug mode install option

Nohzoh:
## Description

Add an option to enable log/debug options after install with unattended script.
Also: instead of installing the centreon-release RPM the repo is installed directly.

## Type of change

## Target serie

- [x] 21.04.x
- [x] 21.10.x (master)

Fresh install from unattended and check install NR.
The debug mode option is by default disabled, it can be enabled by setting a environment variable before launching the unattended script:
`ENV_DEBUG_MODE=true`

Contexts to test:
- with or without the option enabled
- on each supported os

When the option is enabled, log files should appear in /var/log/mysql, /var/log/centreon-broker.

The debug options that can be configured from central UI might be also enabled automatically.
But this part can only be done if the install wizard is played automatically.

I tested it on Centos/7 and OracleLinux/8.


unattended.sh
```diff
@@ -496,16 +495,9 @@ function install_centreon_repo() {
...
-		$PKG_MGR -q install -y $RELEASE_RPM_URL
-		if [ $? -ne 0 ]; then
-			error_and_exit "Could not install Centreon repository"
-		fi
-	else
-		log "INFO" "Centreon repository seems to be already installed"
-	fi
+	yum install -y yum-utils
+	yum-config-manager --add-repo \
+		"https://download.centreon.com/static/centreon-${CENTREON_MAJOR_VERSION}-${EL_MAJOR_VERSION}.repo"
```
quanghungb, line 500: @Nohzoh we may include the repo files directly in the git repo ?

Nohzoh: We can.
The only limitation I see is that the install will rely on Github availability.
It is not much of a risk.

lgcosta: I think if that happens, the script itself will not be available

varinstall/yum-repos/centreon-20.04-el7.repo
```diff
@@ -0,0 +1,63 @@
+[centreon-stable-noarch]
```
sc979, line 1: https://github.com/centreon/centreon/pull/9761/files#diff-2b45a64fb59d94a79d816de404df8c4f30bfccff9cd0b2a94fe4e6d623e2a7f9R7

Are 20.04 versions also supported ?

sc979 (approved):

varinstall/yum-repos/centreon-20.04-el7.repo
Nohzoh: It will be possible to use those files instead of the centreon-release rpm for the 20.04 ; even if this version of the unattended.sh script will not manage this version.

sc979 (approved):

unattended.sh
```diff
@@ -694,6 +686,59 @@ function update_after_installation() {
...
+			INSERT INTO options VALUES ('debug_ldap_import','1');
+			INSERT INTO options VALUES ('debug_sql','1');
+			INSERT INTO options VALUES ('debug_gorgone','1');
+			INSERT INTO options VALUES ('debug_centreontrapd','1');
+EOF
+	fi
+
+	mkdir -p /var/log/mysql
+	chown mysql:mysql /var/log/mysql
+	printf "\n[mariadb]\nlog_output=FILE\ngeneral_log\ngeneral_log_file=/var/log/mysql/mariadb.log\nlog_error\nlog_error=/var/log/mysql/mariadb.err\n" >> /etc/my.cnf.d/custom_log.cnf
```
lgcosta, line 710:
```suggestion
       cat <<EOF >> /etc/my.cnf.d/custom_log.cnf

[mariadb]
log_output=FILE
general_log
general_log_file=/var/log/mysql/mariadb.log
log_error\nlog_error=/var/log/mysql/mariadb.err

EOF
```
Wouldn't using EOF get more intuitive and visual?

unattended.sh
```diff
@@ -694,6 +686,59 @@ function update_after_installation() {
...
+	printf "\nslow_query_log = 1\nslow-query_log_file = /var/log/mysql/slow.log\nlong_query_time = 2" >> /etc/my.cnf.d/centreon.cnf
```
lgcosta, line 711:
```suggestion
        cat <<EOF >> /etc/my.cnf.d/centreon.cnf
slow_query_log = 1
slow-query_log_file = /var/log/mysql/slow.log
long_query_time = 2
EOF
```
here too

lgcosta (changes requested):

PR #9925, opened 2021-06-18
fix(UI): Check pages are fetched before list of allowed pages

Jordan91:
## Description

A message should be displayed “You are not allowed to see this page” for user with no ACL

**Fixes** # (issue)

The message now appears when the user does not have configured Acl access
Capture d’écran 2021-06-18 à 10 29 17

## Type of change

## Target serie

- [x] 20.10.x
- [x] 21.04.x
- [x] 21.10.x (master)

Enable the user account and gives it the ability  just to reach the front end => No Acl .


Thebarda (dismissed):

www/front_src/src/components/ReactRouter/index.tsx
```diff
@@ -2,7 +2,7 @@ import * as React from 'react';
 
 import { connect } from 'react-redux';
 import { Switch, Route, withRouter } from 'react-router-dom';
-import { isEmpty, equals } from 'ramda';
+import { equals } from 'ramda';
```
Thebarda, line 5: `equals` is not used here

Jordan91: And here ? :  line 102 to 106
 (previousProps, nextProps) =>
    equals(previousProps.pages, nextProps.pages) &&
    equals(previousProps.allowedPages, nextProps.allowedPages) &&
    equals(previousProps.externalPagesFetched, nextProps.externalPagesFetched),
);

I just want removed isEmpty, equals was present before my commit

jdelpierre (approved):

Thebarda (approved):

PR #10100, opened 2021-09-09
fix(ui): Move Navigation from centreon-frontend and replace icomoon by material-ui icons

bdauria:
## Description

image (3)

image (2)

## Type of change

## Target serie

- [x] 21.10.x (master)


www/front_src/src/styles/partials/_helpers.scss
```diff
@@ -2,7 +2,7 @@
   margin-bottom: 20px;
 }
 .list-unstyled {
-  list-style: none;
+  // list-style: none;
```
bdauria, line 5:
```suggestion
```

Thebarda (approved):
I hope we will move it to TS, Functional component and remove scss

bdauria:
> I hope we will move it to TS, Functional component and remove scss

This will be a step by step process. This needs a visual polish as well

jdelpierre (approved):

PR #10145, opened 2021-09-17
fix(downtime): fix downtime set with fr locale

kduret:
## Description

fix downtime set with fr locale

**Fixes** IBT-507

## Type of change

## Target serie

- [x] 20.04.x
- [x] 20.10.x
- [x] 21.04.x
- [x] 21.10.x (master)


jeremyjaouen (dismissed):

kduret (commented):
seems it's not enough
Error text is not displayed when downtime start after 2100
Maybe use `addFormRule` instead

kduret:
> seems it's not enough
> Error text is not displayed when downtime start after 2100
> Maybe use `addFormRule` instead

fixed

callapa (approved):

adr-mo (approved):

PR #10167, opened 2021-09-24
fix(TopCounter):  redirect top counter to status detail pages

bdauria:
Description
This allows the User to be redirected to Status details pages when the "Use deprecated pages" flag is activated

Fixes # (issue)

Type of change
 Patch fixing an issue (non-breaking change)

Target serie

 21.10.x (master)


www/front_src/src/Header/StatusCounter/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  const [toggled, setToggled] = React.useState<boolean>();
+
+  const interval = React.useRef<number>();
+
+  const { refreshInterval } = useUserContext();
+
+  const getData = (): void => {
+    axios
+      .get(`./api/${endpoint}`)
+      .then(({ data: retrievedData }) => {
```
Thebarda, line 45: Use getData and useRequest from centreon UI

Thebarda (approved):

Jordan91 (approved):

PR #10246, opened 2021-10-05
chore(npm): Update packages

bdauria:
## Type of change

## Target serie

- [x] 21.10.x (master)


Thebarda (approved):

www/front_src/src/Resources/Listing/columns/State/DetailsTable/Acknowledgement.tsx
```diff
@@ -1,4 +1,6 @@
+/* eslint-disable react/prop-types */
```
jdelpierre, line 1: why this ?

bdauria: I guess a regression in the latest eslint-plugin-react that doesn't handle properly the Props signature. I'll investigate on that later on

jdelpierre: Ok

jdelpierre (approved):

PR #10424, opened 2021-11-22
chore(install): update version to 21.04.9

chgautier:
Update version to 21.04.9 in files


callapa (commented):
Why create an empty PHP file (www/install/php/Update-21.04.9.php) ?

chgautier:
A placeholder/ empty file is required for the upgrade to take place. So we put the php tag and license.

tuntoja (approved):

PR #10495, opened 2021-12-15
enh(stats) : add exception for modules statistics (22.04)

AssanDiomande:
## Description

Add exception for statistics

**Fixes** MON-5760

## Type of change

## Target serie

- [ ] 20.10.x
- [ ] 21.04.x
- [ ] 21.10.x
- [x] 22.04.x (master)

### DIFF
--- a/www/class/centreonStatsModules.class.php
+++ b/www/class/centreonStatsModules.class.php
@@ -125,7 +125,7 @@ class CentreonStatsModules
                 try {
                     $oModuleObject = new $moduleObject();
                     $data[] = $oModuleObject->getStats();
-                } catch (\Exception $e) {
+                } catch (\Throwable $e) {
                     $this->logger->error($e->getMessage, ['context' => $e]);
                 }
             }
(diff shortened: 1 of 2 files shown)


www/class/centreonStatsModules.class.php
```diff
@@ -116,15 +116,18 @@ private function getModuleObjects(array $installedModules)
...
         );
-
         if (is_array($moduleObjects)) {
             foreach ($moduleObjects as $moduleObject) {
-                $oModuleObject = new $moduleObject();
-                $data[] = $oModuleObject->getStats();
+                try {
+                    $oModuleObject = new $moduleObject();
+                    $data[] = $oModuleObject->getStats();
+                } catch (\Throwable $e) {
```
kduret, line 128: You should catch `StatisticsException`

kduret: @callapa let me know what do you think about it :)

AssanDiomande: we don't raise exception, so I think we should leave it only to avoid inheritance problems

kduret: then, let `Throwable` to avoid side effects if an other Exception than `StatisticsException` is thrown

kduret (approved):

jeremyjaouen (approved):

PR #10643, opened 2022-02-01
refactor(header) : Replace Header scss module with material makeStyles(pollerMenu)

Noha-ElAbrouki:
Replace Header scss module with material makeStyles for poller Menu component


www/front_src/src/Header/PollerMenu/index.tsx
```diff
@@ -72,6 +70,21 @@ const useStyles = makeStyles((theme) => ({
   pollerDetailTitle: {
     flexGrow: 1,
   },
+  subMenuToggle: {
+    backgroundColor: '#232f39',
+    boxSizing: 'border-box',
+    display: 'none',
+    left: 0,
+    padding: 10,
```
Thebarda, line 78: Use `spacing` from MUI theme (https://mui.com/customization/default-theme/)

PR #10648, opened 2022-02-01
fix(ui) : kept timeperiod in service reporting page

AssanDiomande:
## Description

kept timeperiod in service reporting page (see last jira ticket comment)

**Fixes** # (issue)

## Type of change

## Target serie

- [ ] 20.10.x
- [ ] 21.04.x
- [x] 21.10.x
- [x] 22.04.x (master)

* Go to Monitoring > Ressource Status
* Click on a server then on "..." (at the right section)
* Select service and a time period then click on "Apply period"
* Check if after the refresh the radio button or period is still selected


www/include/reporting/dashboard/DB-Func.php
```diff
@@ -586,13 +585,20 @@ function getLogInDbForServicesGroup($servicegroupId, $startDate, $endDate, $repo
...
         $serviceGroupStats[$hostServiceid]["SERVICE_DESC"] = $service['service_description'];
 
         if (isset($servicesStats[$hostId][$serviceId])) {
             $serviceGroupStats[$hostServiceid] = $servicesStats[$hostId][$serviceId];
+
+            //Must be here to display link for redirection to a specific server if there is a link between host and server
+            $serviceGroupStats[$hostServiceid]["HOST_ID"] = $hostId;
+            $serviceGroupStats[$hostServiceid]["SERVICE_ID"] = $serviceId;
+            $serviceGroupStats[$hostServiceid]["HOST_NAME"] = $service['host_name'];
+            $serviceGroupStats[$hostServiceid]["SERVICE_DESC"] = $service['service_description'];
```
kduret, line 601: duplicated with lines 589-592

kduret (changes requested):

www/include/reporting/dashboard/DB-Func.php
AssanDiomande: No host and service will appear below the pie chart if I delete lines 498-601

AssanDiomande: fixed

www/include/reporting/dashboard/template/viewServicesLog.ihtml
```diff
@@ -95,7 +93,7 @@
 							<td class='ListColCenter'>{$summary.UNDETERMINED_TF}</td>
 							<td class='ListColRight'>{$summary.UNDETERMINED_TP}%</td>
 							<td class='ListColRight'></td>
-							<td class='ListColRight'>{$summary.UNDETERMINED_A}</td>
```
kduret, line 98: `UNDETERMINED_A` is never set ?

AssanDiomande: Only in viewServiceLog.php line 155 but it is set to null (I had a warning concerning this parameter so I deleted)

jeremyjaouen: I think it should be better to get this value here in getServicesLogs https://github.com/centreon/centreon/blob/develop/www/include/reporting/dashboard/DB-Func.php#L461

In this request, with a sum as it is done for other status.

kduret (approved):

jeremyjaouen (approved):

PR #10658, opened 2022-02-03
refactor(header) : replace header scss module with material make styles service component 

Noha-ElAbrouki:
replace header scss module with material make styles for service component


www/front_src/src/Header/RessourceStatusCounter/Service/index.tsx
```diff
@@ -107,7 +106,7 @@ const ServiceStatusCounter = (): JSX.Element => {
       schema={statusSchema}
     >
       {({ hasPending, data, toggled, toggleDetailedView }): JSX.Element => (
-        <div className={`${styles.wrapper} wrap-right-services`}>
```
jdelpierre, line 110: `wrap-right-services`it is not really used anywhere?

Noha-ElAbrouki: yes , here is an image for this component
serviceComponent

jdelpierre: Yes but I see it has been completely removed, so I don't know if it was already useless before the PR (not used anywhere else but here)

Noha-ElAbrouki: it's the dark mode branch, I didn't  see the effect so i deleted it ?
wrapRightUser

jdelpierre: I agree, it is quite possible that this is obsolete.

jdelpierre (approved):

www/front_src/src/Header/RessourceStatusCounter/index.tsx
```diff
@@ -11,10 +11,28 @@ import { refreshIntervalAtom } from '@centreon/ui-context';
...
+    backgroundColor: theme.palette.common.black,
+    boxSizing: 'border-box',
+    display: 'none',
+    left: theme.spacing(0),
+    padding: theme.spacing(1),
+    position: 'absolute',
+    textAlign: 'left',
+    top: '100%',
+    width: '100%',
+    zIndex: 99,
```
Thebarda, line 28: Is it possible to use the `theme.zIndex.mobileStepper` https://mui.com/customization/default-theme/

bdauria (approved):

Thebarda (approved):

PR #10960, opened 2022-04-11
feat : Mon 12668 switch mode

Noha-ElAbrouki:
create switch mode


www/front_src/src/Header/switchMode/index.tsx
```diff
@@ -0,0 +1,126 @@
+import * as React from 'react';
```
bdauria, line 1: rename the directory to SwitchThemeMode as well as the Component

bdauria, line 1: rename the directory to SwitchThemeMode as well as the Component

bdauria (approved):

Thebarda (approved):

PR #10972, opened 2022-04-13
 refactor: MON-12744-useClsx

Noha-ElAbrouki:
replace classnames with clsx


bdauria:
@Noha-ElAbrouki You need to change to base to the dark-mode base to avoid future conflicts (some classnames uses already have been cleaned up there)

bdauria:
@Noha-ElAbrouki You need to change to base to the dark-mode base to avoid future conflicts (some classnames uses already have been cleaned up there)

PR #10990, opened 2022-04-19
enh(core): Upgrade to react 18 🎉

Thebarda:
## Description

🎉🎉🎉 This upgrades our code to React 18 🎉🎉🎉
It also remove the import of React by change how we emit the JSX while compiling it with the TypeScript compiler

## Type of change

- [ ] Patch fixing an issue (non-breaking change)
- [ ] New functionality (non-breaking change)
- [ ] Breaking change (patch or feature) that might cause side effects breaking part of the Software
- [ ] Updating documentation (missing information, typo...)

## Target serie

- [ ] 20.10.x
- [ ] 21.04.x
- [ ] 21.10.x
- [x] 22.04.x (master)

<h2> How this pull request can be tested ? </h2>

Everything should work as before

## Checklist

- [x] I have followed the **coding style guidelines** provided by Centreon
- [ ] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [ ] I have commented my code, especially **hard-to-understand areas** of the PR.
- [ ] I have made corresponding changes to the **documentation**.
- [x] I have **rebased** my development branch on the base branch (master, maintenance).

PR #10991, opened 2022-04-19
fix(App/index): Mon 13034 white background displayed during pages navigation  when the theme mode is dark

Noha-ElAbrouki:
fix background color for displayed pages when the theme mode is dark


bdauria (approved):

Thebarda (approved):
That's not the correct way to fix this but it's ok for the moment. [Here](https://github.com/centreon/centreon/pull/10990/commits/fed1c673dbb252841a9675d8599f0dbf68bfd9f3) is the real fix

PR #11057, opened 2022-05-03
fix(apache): manage custom uri

kduret:
## Description

manage apache custom uri

**Fixes** MON-13189

## Type of change

## Target serie

- [ ] 20.10.x
- [ ] 21.04.x
- [ ] 21.10.x
- [x] 22.04.x (master)


www/index.php
```diff
@@ -34,28 +34,16 @@
...
+require_once __DIR__ . "/class/centreonAuth.class.php";
 require_once __DIR__ . "/class/centreonLog.class.php";
 require_once __DIR__ . "/class/centreonDB.class.php";
 
-/*
- * Get auth type
- */
-global $pearDB;
-$pearDB = new CentreonDB();
+const AUTOLOGIN_FIELDS = ['autologin' , 'useralias', 'token'];
```
kduret, line 43:
```suggestion
const AUTOLOGIN_FIELDS = ['autologin', 'useralias', 'token'];
```

www/index.php
```diff
@@ -75,139 +63,162 @@
...
-        $argP = filter_var(
-            $_POST['p'] ?? $_GET["p"] ?? null,
-            FILTER_VALIDATE_INT
+    return;
+}
+
+global $pearDB;
+$pearDB = new CentreonDB();
+
+$dbResult = $pearDB->query("SELECT `value` FROM `options` WHERE `key` = 'enable_autologin'");
```
kduret, line 72: query only if autologin arguments are set

a-launois (dismissed):

adr-mo (dismissed):

jeremyjaouen (dismissed):

a-launois (dismissed):

adr-mo (dismissed):

www/index.php
```diff
@@ -75,139 +64,172 @@
...
-            header("Location: " . $headerRedirection);
+        }
+    } elseif (isset($centreon->user->default_page) && $centreon->user->default_page != '') {
+        // get more details about the default page
+        $stmt = $pearDB->prepare(
+            "SELECT topology_url, is_react FROM topology WHERE topology_page = ? LIMIT 0, 1"
+        );
+        $pearDB->execute($stmt, [$centreon->user->default_page]);
+
+        if ($stmt->rowCount() && ($topologyData = $stmt->fetch()) && $topologyData['is_react']) {
```
callapa, line 222: Although it is functional, any value (!= 0) in is_react will validate the test.
It would be better to verify either that the value is different from 0 or equal to 1

callapa (approved):

PR #11158, opened 2022-05-24
ft(UI): MON-13458 frontend - Add card in details panel like groupChips

Jordan91:
## Description

MON-13458 frontend - Add card in details panel like groupChips

## Type of change

## Target serie

- [x] 22.10.x


www/front_src/src/Resources/Details/tabs/Details/DetailsCard/DetailsChips.tsx
```diff
@@ -1,18 +1,17 @@
...
-import { ResourceType } from '../../../../models';
+import { Category, Group } from '../../../models';
 
-import GroupChip from './GroupChip';
+import DetailsChip from './DetailsChip';
 
 interface Props {
-  details: ResourceDetails | undefined;
+  getType: () => CriteriaNames;
+  metaResources?: Array<Category | Group>;
```
bdauria, line 13: I don't think metaResource is the proper name for the entities as it can be confused with meta services

bdauria:
@Jordan91 it seems like there are too many files in comparison to the work that must be integrated. It might be because the base branch is not properly synchronized with develop

www/front_src/src/Resources/Details/tabs/Details/DetailsCard/DetailsChips.tsx
Thebarda: > I don't think metaResource is the proper name for the entities as it can be confused with meta services

I think we can rename it to `chips`

Thebarda:
I think you have unwanted changed files

Jordan91:
Other PR open

PR #11244, opened 2022-06-15
fix(php): manage compatibility 22.04 with extensions

kduret:
## Description

MBI compatibility with centreon web 22.04 was broken due to namespace renaming
this adds compatibility classes which will be only on 22.04.x
for 22.10, MBI, autodisco & BAM will be update to manage this change.

**Fixes** MON-13975

## Type of change

## Target serie

- [ ] 21.04.x
- [ ] 21.10.x
- [x] 22.04.x
- [ ] 22.10.x (master)


adr-mo (approved):

kduret (commented):
add `@depreacted` in class php doc and describe why

adr-mo (approved):

jeremyjaouen (approved):

PR #11286, opened 2022-06-24
fix(web): display command with status$ in the command definition

s-duret:
## Description

In the check command definition we have to define some options like this:
`--interface='^$_SERVICEINTERFACENAME$$$'`

It allows the user to only monitor the resource he defined in the macro value.
But Centreon-Engine will replace the last two $$ by only one $.

We have to deal with it.

**Fixes** MON-14128

## Type of change

## Target serie

- [ ] 21.04.x
- [x] 21.10.x
- [x] 22.04.x
- [x] 22.10.x (master)

### DIFF
--- a/src/Centreon/Domain/Monitoring/CommandLineTrait.php
+++ b/src/Centreon/Domain/Monitoring/CommandLineTrait.php
@@ -45,7 +45,7 @@ trait CommandLineTrait
     ): string {
         // if the command line contains $$ after a macro (so $$$), delete one of them to match with
         // the command executed by centreon-engine
-        $configurationCommand = str_replace('$$$', '$$', $configurationCommand);
+        $configurationCommand = str_replace('$$', '$', $configurationCommand);
         $macroPasswordNames = [];
         foreach ($macros as $macro) {
             if ($macro->isPassword()) {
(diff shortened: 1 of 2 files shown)


jeremyjaouen (dismissed):

src/Centreon/Domain/Monitoring/CommandLineTrait.php
```diff
@@ -43,6 +43,9 @@ private function buildCommandLineFromConfiguration(
         array $macros,
         string $replacementValue
     ): string {
+        // if the command line contains $$ after a macro (so $$$), delete one of them to match with
+        // the command executed by centreon-engine
+        $configurationCommand = str_replace('$$$', '$$', $configurationCommand);
```
kduret, line 48: I think centreon-engine needs `$$` to escape `$`, so this replacement should be enough to fix the issue
```suggestion
        $configurationCommand = str_replace('$$', '$', $configurationCommand);
```

s-duret: The function _buildCommandLineFromConfiguration_ isn't used to generate the configuration. It's used for the monitoring page in order to display the command line.
A lot of plugins packs contain this kind of option:
`--interface='^$_SERVICEINTERFACENAME$$$'`
Because, you're right, centreon-engine needs $$ to execute $
So I wanted to fix this specific case

s-duret: following our discussion, it's better to replace $$ by $

jeremyjaouen (dismissed):

kduret (dismissed):

wtermellil (dismissed):

callapa (dismissed):

jeremyjaouen (dismissed):

TamazC (approved):

callapa (approved):

PR #11317, opened 2022-06-30
fix(widget): add empty value for poller filter

s-duret:
## Description

Fix the poller filter in the widget preferences.
We have to let the user select an empty line.

**Fixes** MON-14174

## Type of change

## Target serie

- [x] 21.04.x
- [x] 21.10.x
- [x] 22.04.x
- [x] 22.10.x (master)

### DIFF
--- a/www/class/centreonWidget/Params/Connector/Poller.class.php
+++ b/www/class/centreonWidget/Params/Connector/Poller.class.php
@@ -89,6 +89,7 @@ class CentreonWidgetParamsConnectorPoller extends CentreonWidgetParamsList
                     );
                 }
 
+                $tab = array(null => null);
                 while (($record = $statement->fetch(\PDO::FETCH_ASSOC)) !== false) {
                     $tab[$record['id']] = $record['name'];
                 }
--- a/www/class/centreonWidget/Params/Connector/Poller.class.php
+++ b/www/class/centreonWidget/Params/Connector/Poller.class.php
@@ -89,7 +89,7 @@ class CentreonWidgetParamsConnectorPoller extends CentreonWidgetParamsList
                     );
                 }
 
-                $tab = [];
+                $tab = [null => null];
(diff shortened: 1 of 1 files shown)


www/class/centreonWidget/Params/Connector/Poller.class.php
```diff
@@ -89,6 +89,7 @@ public function getListValues($paramId)
                     );
                 }
 
+                $tab = array(null => null);
```
jeremyjaouen:
```suggestion
                $tab = [];
```
Could you check with an empty array if this correct the bug ?

jeremyjaouen (dismissed):

wtermellil (dismissed):

dmyios (dismissed):

www/class/centreonWidget/Params/Connector/Poller.class.php
```diff
@@ -89,6 +89,7 @@ public function getListValues($paramId)
                     );
                 }
 
+                $tab = [];
```
tanguyvda: doesn't work on 21.10 at least 

```suggestion
                $tab = [null => null];
```

TamazC:
The issue is already fix within https://github.com/centreon/centreon/commit/1c7dbfa07170a28d94446ca2547371826ba2ab16

@s-duret could you please check the fix and close it if OK ?

Thanks

s-duret:
> The issue is already fix within [1c7dbfa](https://github.com/centreon/centreon/commit/1c7dbfa07170a28d94446ca2547371826ba2ab16)
> 
> @s-duret could you please check the fix and close it if OK ?
> 
> Thanks

Hello @TamazC 
Yes you're right, it the same fix.
I don't understand, my PR was done in june and the fix you gave me was done 2 weeks ago.
No matter, I will close this issue.

PR #11330, opened 2022-07-04
fix(web): enable the deletion of text parameters on contacts

s-duret:
## Description

It was impossible to undefined contact parameters when they were already defined.

**Fixes** MON-14206

## Type of change

## Target serie

- [x] 21.04.x
- [x] 21.10.x
- [x] 22.04.x
- [x] 22.10.x (master)

### DIFF
--- a/www/include/configuration/configObject/contact/DB-Func.php
+++ b/www/include/configuration/configObject/contact/DB-Func.php
@@ -694,7 +694,6 @@ function updateContact_MC($contact_id = null)
         return;
     }
 
-    $ret = array();
     $ret = $form->getSubmitValues();
 
     // Remove all parameters that have an empty value in order to keep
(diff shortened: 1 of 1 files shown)


jeremyjaouen (dismissed):

wtermellil (dismissed):

dmyios (dismissed):

www/include/configuration/configObject/contact/DB-Func.php
```diff
@@ -696,6 +696,15 @@ function updateContact_MC($contact_id = null)
 
     $ret = array();
```
TamazC:
```suggestion
```

TamazC (dismissed):

jeremyjaouen (approved):

adr-mo (approved):

PR #11393, opened 2022-07-18
enh(Header/userMenu):reduce spacing user menu

Noha-ElAbrouki:
reduce spacing and update style of userMenu


www/front_src/src/Header/SwitchThemeMode/index.tsx
```diff
@@ -1,105 +1,70 @@
...
-    !equals(prop, 'darkModeSvg'),
-})<StyleProps>(({ theme, darkModeSvg, lightModeSvg }) => ({
-  '& .MuiSwitch-switchBase': {
+const useStyles = makeStyles((theme) => ({
+  container: {
+    '& .MuiSwitch-thumb': {
+      backgroundColor: 'white',
+    },
+    '& .MuiSwitch-track': {
+      backgroundColor: '#aab4be',
```
jdelpierre, line 19: not existing a color on the theme ?

Noha-ElAbrouki: because I want a fix color either for light or dark because the header is black in both modes

Thebarda (approved):

jdelpierre (approved):

Issue #11480, opened 2022-08-02
Illegal characters in contact groups imported from LDAP


proxyconcept:
# BUG REPORT INFORMATION

### Prerequisites

***Versions***

centreon-web-22.04.1-8.el8.noarch

***Operating System***

Red Hat Enterprise Linux 8.6 (Ootpa)

### Description

With LDAP backend, the imported contact groups are stored in the database without any check on the name.

This is not a real problem for Centreon Web: administration, synchronisation and authentication are working.
But if a contact group is used for notification, we can have a problem with illegal characters in the nagios configuration files generated.

### Steps to Reproduce

1. Have a group in your LDAP server with a name containing a nagios illegal character (by default `~!$%^&*"|'<>?,()=`)
2. Configure the LDAP backend in Centreon (with group support) and synchronize (LDAP groups imported as contact groups)
3. Edit a host and link for notifications the contact group from LDAP (with illegal chars)
4. Generate the nagios configuration files for the poller

### Describe the received result

Generating files... NOK
Aborted.
```
Error The name of contact group 'xxx (zzz)' contains one or more illegal characters.
Error Cannot resolve contact group xxx (zzz)'
```

The generated files in `/var/cache/centreon/config/engine/1`  contains the contact group names with illegal characters.

### Describe the expected result

Generating files... OK

The generated files in `/var/cache/centreon/config/engine/1`  should contains valid contact group names (with illegal characters removed).

### Additional relevant information

It's a real problem because must be able to interconnect with the LDAP server and it's not possible to impose the naming of the LDAP groups.

proxyconcept:
There is no problem with "normal" contact groups (created by centreon web or clapi) because they are sanitized before being saved (see #9220).

But nothing similar is doing when groups are imported from the LDAP backends.

At first I considered adding same sanitize function when contact groups are created from LDAP (before being saved)... But it seems difficult to store a name different from the original: there is a lot of impact if we don't want to break synchronisation process (it's necessary to verify the existence of a group in DB and in LDAP by it's name).

Then I gave up this input filtering, since the main problem lies in the output (nagios configuration files generation).

PR #11689, opened 2022-09-05
fix(conf): Sanitize contact group names from illegal characters (#11480)

proxyconcept:
Contact groups name can contains illegal characters when imported from LDAP
(sanitize the input is complex due to synchronization process).
Therefore illegal characters are removed only when generating nagios configuration files
(and so according to the current poller configuration for illegal characters).

Refs: #11480
Signed-off-by: Grégory Marigot <gmarigot@teicee.com>

## Description

Contact groups name can contains illegal characters when imported from LDAP
(sanitize the input is complex due to synchronization process).
Therefore illegal characters are removed only when generating nagios configuration files
(and so according to the current poller configuration for illegal characters).

**Fixes** #11480

## Type of change

## Target serie

- [ ] 21.04.x
- [ ] 21.10.x
- [x] 22.04.x
- [x] 22.10.x (master)

1. Have a group in your LDAP server with a name containing a nagios illegal character (by default `~!$%^&*"|'<>?,()=`)
2. Configure the LDAP backend in Centreon (with group support) and synchronize (LDAP groups imported as contact groups)
3. Edit a host and link for notifications the contact group from LDAP (with illegal chars)
4. Generate the nagios configuration files for the poller

The generation should be OK and the nagios configuration files must contains the sanitized name for contact groups0

- [x] I have followed the **coding style guidelines** provided by Centreon
- [x] I have commented my code, especially new **classes**, **functions** or any **legacy code** modified. (***docblock***)
- [x] I have commented my code, especially **hard-to-understand areas** of the PR.

### DIFF
--- a/www/class/config-generate-remote/ContactGroup.php
+++ b/www/class/config-generate-remote/ContactGroup.php
@@ -48,6 +48,8 @@ class ContactGroup extends AbstractObject
     protected $stmtContact = null;
     protected $stmtCgService = null;
 
+    protected $illegalCharsTab = array();
+
     /**
      * Generate contact group cache
      *
@@ -184,7 +186,34 @@ class ContactGroup extends AbstractObject
     }
 
     /**
-     * Generate contact group and get contact group name
+     * Check illegal char defined into nagios.cfg file
+     *
+     * @param string $name The string to sanitize
+     * @return string The string sanitized
+     */
+    protected function checkIllegalChar($name)
+    {
+        $pollerId = $this->backendInstance->getPollerId();
+
+        if (! isset($this->illegalCharsTab[ $pollerId ])) {
+            $query = "SELECT illegal_object_name_chars FROM cfg_nagios WHERE nagios_server_id = :pollerId";
+            $stmt = $this->backendInstance->db->prepare($query);
+            $stmt->bindParam(':pollerId', $pollerId, PDO::PARAM_INT);
+            $stmt->execute();
+
+            $illegalCharsStr = implode('', $stmt->fetchAll(PDO::FETCH_COLUMN));
+            $illegalCharsStr = html_entity_decode($illegalCharsStr, ENT_QUOTES, "UTF-8");
+            $this->illegalCharsTab[ $pollerId ] = array_unique(str_split($illegalCharsStr));
+        }
+
+        if (! empty($this->illegalCharsTab[ $pollerId ])) {
+            $name = str_replace($this->illegalCharsTab[ $pollerId ], '_', $name);
+        }
+        return $name;
+    }
+
+    /**
+     * Generate contact group and get contact group name (sanitized)
      *
      * @param null|integer $cgId
      * @return void|string
@@ -212,6 +241,8 @@ class ContactGroup extends AbstractObject
         if ($this->checkGenerate($cgId)) {
             return $this->cg[$cgId]['cg_name'];
         }
+
+        $this->cg[$cgId]['cg_name'] = $this->checkIllegalChar($this->cg[$cgId]['cg_name']);
 
         $this->getContactFromCgId($cgId);
 
--- a/www/class/config-generate/contactgroup.class.php
+++ b/www/class/config-generate/contactgroup.class.php
@@ -58,6 +58,8 @@ class Contactgroup extends AbstractObject
     protected $stmt_contact = null;
     protected $stmt_cg_service = null;
 
+    protected $illegalCharsTab = array();
+
     protected function getCgCache()
     {
         $stmt = $this->backend_instance->db->prepare("SELECT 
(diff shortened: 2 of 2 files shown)


www/class/config-generate-remote/ContactGroup.php
```diff
@@ -48,6 +48,8 @@ class ContactGroup extends AbstractObject
     protected $stmtContact = null;
     protected $stmtCgService = null;
 
+    protected $illegalCharsTab = array();
```
TamazC, line 51:
```suggestion
```

TamazC: There are no need of this property.
The logic of sinitazing could be put in a method.

www/class/config-generate-remote/ContactGroup.php
```diff
@@ -184,7 +186,34 @@ public function getContactFromCgId(int $cgId)
...
+        }
+
+        if (! empty($this->illegalCharsTab[ $pollerId ])) {
+            $name = str_replace($this->illegalCharsTab[ $pollerId ], '_', $name);
+        }
+        return $name;
+    }
+
+    /**
+     * Generate contact group and get contact group name (sanitized)
```
TamazC, line 216:
```suggestion
    private function sanitizeNameFromIllegalChars(string $name): string
    {
        $pollerId = $this->backendInstance->getPollerId();

        $illegalChars = $this->getIllegalCharsByNagiosServerId($pollerId);

        if (empty($illegalChars)) {
            return $name;
        }

        return str_replace($illegalChars, '_', $name);
    }

    private function getIllegalCharsByNagiosServerId(int $nagiosServerId): array
    {
        $query = 'SELECT illegal_object_name_chars FROM cfg_nagios WHERE nagios_server_id = :nagiosServerId';
        $stmt = $this->backendInstance->db->prepare($query);
        $stmt->bindParam(':nagiosServerId', $nagiosServerId, PDO::PARAM_INT);
        $stmt->execute();

        $illegalCharsStr = implode('', $stmt->fetchAll(PDO::FETCH_COLUMN));
        $illegalCharsStr = html_entity_decode($illegalCharsStr, ENT_QUOTES, 'UTF-8');

        return array_unique(str_split($illegalCharsStr));
    }
```

TamazC: The name checkIllegalChar  doesn't describe what the method is actually doing. This method retreieves illegal chars and sanitizes the string and returns them.

I propose to split it in 2 methods : a get method (from database) and a sinitize method. Then the sinitize method calls get method.

I'd make them private. no need of protected.

www/class/config-generate-remote/ContactGroup.php
```diff
@@ -213,6 +242,8 @@ public function generateFromCgId(?int $cgId)
             return $this->cg[$cgId]['cg_name'];
         }
 
+        $this->cg[$cgId]['cg_name'] = $this->checkIllegalChar($this->cg[$cgId]['cg_name']);
```
TamazC, line 245:
```suggestion
        $this->sanitizeNameFromIllegalChars($this->cg[$cgId]['cg_name']);
```

www/class/config-generate/contactgroup.class.php
```diff
@@ -58,6 +58,8 @@ class Contactgroup extends AbstractObject
     protected $stmt_contact = null;
     protected $stmt_cg_service = null;
 
+    protected $illegalCharsTab = array();
```
TamazC, line 61: Same comments here

www/class/config-generate/contactgroup.class.php
```diff
@@ -187,6 +189,36 @@ public function getContactFromCgId(int $cgId): void
     }
 
     /**
+     * Check illegal char defined into nagios.cfg file
+     *
+     * @param string $name The string to sanitize
+     * @return string The string sanitized
+     */
+    protected function checkIllegalChar($name)
```
TamazC, line 197: And here

www/class/config-generate-remote/ContactGroup.php
proxyconcept: > The name checkIllegalChar doesn't describe what the method is actually doing. This method retreieves illegal chars and sanitizes the string and returns them.
> 
> I propose to split it in 2 methods : a get method (from database) and a sinitize method. Then the sinitize method calls get method.
> 
> I'd make them private. no need of protected.

I'm agree with you...
If I choose a unique function named "checkIllegalChar", it's only because it's already like that in other existing files:

- www/class/centreon.class.php:291
- www/class/centreonHost.class.php:618
- www/class/centreonService.class.php:307
- www/class/centreon-clapi/centreonObject.class.php:638

So I tried to stay consistent with the existing code, I don't know which is preferable.

www/class/config-generate-remote/ContactGroup.php
proxyconcept: > There are no need of this property. The logic of sinitazing could be put in a method.

I added this property to have a cache of the SQL request, in order to avoid querying the database for each name to sanitize (especially since this process is not on the input of a unique value, but when the entire engine configuration is generated).

PR #12005, opened 2022-10-17
enh(ResourcesListing):Use upper and lower case typography allows better reading/distinction and understanding information

Noha-ElAbrouki:
use lowerCase for status name of resources status 

**screens**

resoursoo

### DIFF
--- a/www/front_src/src/Resources/Listing/columns/Status.tsx
+++ b/www/front_src/src/Resources/Listing/columns/Status.tsx
@@ -129,13 +129,16 @@ const StatusColumn = ({
 
     const statusName = row.status.name;
 
+    const lowerLetter = (name: string): string =>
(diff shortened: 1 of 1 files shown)


www/front_src/src/Resources/Listing/columns/Status.tsx
```diff
@@ -129,13 +129,16 @@ const StatusColumn = ({
...
+      name.charAt(0) + name.slice(1).toLocaleLowerCase();
+
     return (
       <div className={classes.statusColumn}>
         {isHovered ? (
           <StatusColumnOnHover actions={actions} row={row} />
         ) : (
           <StatusChip
-            label={t(statusName)}
+            label={lowerLetter(t(statusName))}
```
Thebarda, line 141: This must be done in the StatusChip component in centreon-frontend

Thebarda (approved):

